PR #23, opened 2025-01-11
support python < 3.9 & fix nan when pytorch < 2.5 & fix deepspeed train type mismatch

conansherry:
1. support python < 3.9
    change type list to List
2. fix nan when pytorch < 2.5
    see [https://github.com/huggingface/diffusers/commit/01bd79649e0bc01bd3de48d6829a6d9514a361a5](https://github.com/huggingface/diffusers/commit/01bd79649e0bc01bd3de48d6829a6d9514a361a5)
3. fix deepspeed train type mismatch
original code:
`context_aware_representations = self.c_embedder(context_aware_representations)`
fix code:
`context_aware_representations = self.c_embedder(context_aware_representations.to(dtype=x.dtype))`

### DIFF
--- a/hunyuan_model/models.py
+++ b/hunyuan_model/models.py
@@ -756,11 +756,11 @@ class HYVideoDiffusionTransformer(nn.Module):  # ModelMixin, ConfigMixin):
         if self.attn_mode == "torch" and not self.split_attn:
             # initialize attention mask: bool tensor for sdpa, (b, 1, n, n)
             bs = img.shape[0]
-            attn_mask = torch.zeros((bs, 1, max_seqlen_q, max_seqlen_q), dtype=torch.bool, device=text_mask.device)
+            attn_mask = torch.zeros((bs, 1, max_seqlen_q), dtype=torch.bool, device=text_mask.device)
 
             # set attention mask with total_len
             for i in range(bs):
-                attn_mask[i, :, : total_len[i], : total_len[i]] = True
+                attn_mask[i, :, : total_len[i]] = True
             total_len = None  # means we don't use split_attn
 
         freqs_cis = (freqs_cos, freqs_sin) if freqs_cos is not None else None
--- a/cache_latents.py
+++ b/cache_latents.py
@@ -1,6 +1,6 @@
 import argparse
 import os
-from typing import Optional, Union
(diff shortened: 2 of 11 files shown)


conansherry:
> Thank you for this, I will merge this. However, please note that it is difficult to maintain compatibility in the future (and compatibility may be lost accidentally), because we do not have PyTorch 3.9 environment. If that case occurs, please send another PR to fix it.
> 
> Regarding 2, there is a comment in the diffusers' pull request, will it be okay with PyTorch 2.5.1? : [huggingface/diffusers@01bd796#commitcomment-151162702](https://github.com/huggingface/diffusers/commit/01bd79649e0bc01bd3de48d6829a6d9514a361a5#commitcomment-151162702)

yes, pytorch 2.5.1 is okay.

hunyuan_model/models.py
```diff
@@ -756,11 +756,11 @@ def forward(
         if self.attn_mode == "torch" and not self.split_attn:
             # initialize attention mask: bool tensor for sdpa, (b, 1, n, n)
             bs = img.shape[0]
-            attn_mask = torch.zeros((bs, 1, max_seqlen_q, max_seqlen_q), dtype=torch.bool, device=text_mask.device)
+            attn_mask = torch.zeros((bs, 1, max_seqlen_q), dtype=torch.bool, device=text_mask.device)
 
             # set attention mask with total_len
             for i in range(bs):
-                attn_mask[i, :, : total_len[i], : total_len[i]] = True
+                attn_mask[i, :, : total_len[i]] = True
```
kohya-ss, line 763: This fix seems to result in the following error in PyTorch 2.5.1:
```
x = F.scaled_dot_product_attention(q, k, v, attn_mask=attn_mask, dropout_p=drop_rate, is_causal=causal)
RuntimeError: The expanded size of the tensor (24) must match the existing size (3) at non-singleton dimension 1.  Target sizes: [3, 24, 2296, 2296].  Tensor sizes: [3, 1, 2296]
```

Could you please revert this fix? Without this fix, it would work with `--split_attn` on versions prior to PyTorch 2.5.1.

bmaltais:
With the introduction of uv as the python module manager, is supporting python 3.9 really necessary? Why support such an old version?

PR #25, opened 2025-01-11
Require torch >= 2.5.1 for sdpa

nivibilla:
According to https://github.com/kijai/ComfyUI-HunyuanVideoWrapper/issues/126 and also my own testing. Torch 2.5.1 is required to not get black videos when using sdpa attn_mode.


nivibilla:
Yep makes sense. Thanks.

PR #28, opened 2025-01-12
Sample images during training

kohya-ss:
- Enable prompt options (like `--w`).
- Common some functions for generate and train.


sdbds:
image
According to the Hunyuan video paper, fs should be linear, so if the number of inference steps is 20, then fs should be 14.5.

PR #31, opened 2025-01-12
Add CAME optimizer, a confidence guided, memory efficient optimizer

Sarania:
This simple PR adds the CAME optimizer ( https://github.com/yangluo7/CAME/ ) following previous examples in musubi-tuner and sdscripts. CAME is a confidence guided, memory efficient optimizer that 's useful in my tests. Default optimizer args that I use are:

```
--optimizer_args weight_decay=0.01 eps=(1e-30,1e-16) betas=(0.9,0.999,0.9999)
```


Sarania:
Ahh I see, I knew that worked for some optimizers (at least with sdscripts) but I didn't know it worked for all of them. No worries then!

PR #33, opened 2025-01-13
add lycoris support

sdbds:
add lycoris support


hv_train_network.py
```diff
@@ -1216,20 +1216,32 @@ def train(self, args):
...
+                network = network_module.create_network_hunyuan_video(
+                    1.0,
+                    args.network_dim,
+                    args.network_alpha,
+                    vae,
+                    None,
+                    transformer,
+                    neuron_dropout=args.network_dropout,
+                    **net_kwargs,
+                )
```
kohya-ss: I think it would be more robust to judge whether it has a function `create_network_hunyuan_video` rather than whether it has the name `lycoris`.

hv_train_network.py
```diff
@@ -1216,20 +1216,32 @@ def train(self, args):
...
         if network is None:
             return
 
-        network.prepare_network(args)
+        if not network_module.__name__.startswith("lycoris"):
```
kohya-ss: I think it's a good idea to call this function only if it exists for this too.

sdbds:
update

Enyakk:
Possible support appreciated on Lycoris LoKr conversion issue:
[https://github.com/kohya-ss/musubi-tuner/issues/68](url)

PR #34, opened 2025-01-13
Add option to specify fps in generate command

nivibilla:
Hi, very small change. Im training my videos with 12fps due to memory constraints so it would be nice to be able to specify that at generation too.


nivibilla:
Yes, that is Intentional because I trimmed my data to 5 secs in 12fps. But when training it's only 60 frames. When when inferencing due to default 24fps it was showing in fast motion putting it down to 12fps made it seem normal if that makes sense because the model learnt to show 5 secs of video in 60 frames

PR #37, opened 2025-01-14
Add LoRA Merger script, Updates README files.

kaykyr:


jurgenprins:
Would it be possible to extend/modify this scripts so that it just saves the merged loras into single lora file (not merge with base model) , just like sd-scripts/networks/merge_lora.py does?

PR #40, opened 2025-01-15
Fix some bugs of full fine turning.

wangqiang9:
1. save model to output_dir.
2. get freqs_cos and freqs_sin when parallel train.

### DIFF
--- a/hv_train.py
+++ b/hv_train.py
@@ -972,7 +972,7 @@ class FineTuningTrainer:
                 is_lora=False,
             )
 
-            save_file(transformer.state_dict(), ckpt_name, sai_metadata)
+            save_file(transformer.state_dict(), ckpt_file, sai_metadata)
             if args.huggingface_repo_id is not None:
                 huggingface_utils.upload(args, ckpt_file, "/" + ckpt_name, force_sync_upload=force_sync_upload)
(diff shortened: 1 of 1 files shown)


hv_train.py
```diff
@@ -1033,7 +1033,10 @@ def remove_model(old_ckpt_name):
 
                     pos_emb_shape = latents.shape[1:]
                     if pos_emb_shape not in pos_embed_cache:
-                        freqs_cos, freqs_sin = get_rotary_pos_embed_by_shape(transformer, latents.shape[2:])
+                        if isinstance(transformer, torch.nn.parallel.DistributedDataParallel):
+                            freqs_cos, freqs_sin = get_rotary_pos_embed_by_shape(transformer.module, latents.shape[2:])
+                        else:
+                            freqs_cos, freqs_sin = get_rotary_pos_embed_by_shape(transformer, latents.shape[2:])
```
kohya-ss: Is it possible to use `accelerator.unwrap_model` instead of `isinstance`?

wangqiang9: yeap. In my testing, it will report an error

wangqiang9: yeap, done.

PR #63, opened 2025-01-26
Add Validation Splits and Logging, Refactor Dataset Blueprints, and Improve Float8 Support

NSFW-API:
• Introduces separate train/val dataset groups in the blueprint config and uses them for both cache_latents.py and cache_text_encoder_outputs.py.  
• Extends hv_train_network.py with a validate() function that runs on the val_dataset_group each epoch, computes MSE loss, and logs val_loss via Accelerate (e.g., accelerator.log(...)).  
• Refactors config_utils.py to return separate train_dataset_group and val_dataset_group, combining them when needed (all_datasets).  
• Adds optional float8 fallback handling in token_refiner.py, safely casting float8 → float for calculations and then back to float8.  
• Adjusts cache skipping/keeping logic to handle old cache files, and changes the code to enumerate all_datasets instead of only train datasets.  
• Overall, this PR makes it possible to run a distinct validation pass each epoch, log validation performance, and unify caching for both train and val.

Screenshot 2025-01-26 093517

### DIFF
--- a/cache_text_encoder_outputs.py
+++ b/cache_text_encoder_outputs.py
@@ -5,17 +5,15 @@ from typing import Optional, Union
 import numpy as np
 import torch
 from tqdm import tqdm
+import accelerate
 
 from dataset import config_utils
 from dataset.config_utils import BlueprintGenerator, ConfigSanitizer
-import accelerate
-
 from dataset.image_video_dataset import ItemInfo, save_text_encoder_output_cache
 from hunyuan_model import text_encoder as text_encoder_module
 from hunyuan_model.text_encoder import TextEncoder
 
 import logging
-
 from utils.model_utils import str_to_dtype
 
 logger = logging.getLogger(__name__)
@@ -23,7 +21,7 @@ logging.basicConfig(level=logging.INFO)
 
 
 def encode_prompt(text_encoder: TextEncoder, prompt: Union[str, list[str]]):
-    data_type = "video"  # video only, image is not supported
+    data_type = "video"  # typically 'video'; can be changed if needed
     text_inputs = text_encoder.text2tokens(prompt, data_type=data_type)
 
     with torch.no_grad():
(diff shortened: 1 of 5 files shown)


cache_text_encoder_outputs.py
NSFW-API, line 1: Most of this file wasn't intended to change, `o1` just went a bit overboard. Will cut it down to only necessary changes.

kohya-ss: Thank you, that's very helpful, it takes me some extra time to understand the changes.

NSFW-API: @kohya-ss Should be ready for you to take a look.

hunyuan_model/token_refiner.py
```diff
@@ -233,8 +233,18 @@ def forward(
         if mask is None:
             context_aware_representations = x.mean(dim=1)
         else:
-            mask_float = mask.float().unsqueeze(-1)  # [b, s1, 1]
```
NSFW-API, line 236: I was getting a crash on my 4090 about type casting I think, this resolved it.

NSFW-API:
> Thank you! I think this is a great starting point.
> 
> As for the dataset settings, just an idea, how about adding the `is_val` (or `is_validation`) attribute to the dataset? This might help minimize modifications to cache_latents.py etc.

I missed this earlier. Will take a look at what it would involve but I don't see why that wouldn't work.

NSFW-API:
I think I need to apply the same stable loss to the normal loss graph too, not just the validation loss.

NSFW-API:
> Thank you! I think this is a great starting point.
> 
> As for the dataset settings, just an idea, how about adding the `is_val` (or `is_validation`) attribute to the dataset? This might help minimize modifications to cache_latents.py etc.

I've implemented this change, no longer needing to modify the caching scripts. Good call out!

NSFW-API:
After re-testing, looks like the cache_latents.py script is failing. It might need some changes to accommodate the `is_val` change after all. Will resubmit once I make sure all parts are working as expected.

NSFW-API:
Ok, just needed a small change to those caching files.

Enyakk:
I am trying out the PR and it looks like great instrument to add to the toolkit. However two changes, if practical, would enhance the effectiveness:
- Large datasets might only have 1-2 epochs during the entirety of the training run so a steps_validate option would be appreciated
- Multiple named validation datasets in tensorboard: Training contents may have multiple goals. Being able to measure each goal against a defined validation dataset (ie. character A, concept B, regularization C) and display the current validation state would enhance productivity even further.

NSFW-API:
I got reports that this isn't working when you don't provide a validation dataset, which it was previously as validation should of course be optional. Will see why that broke with the latest commit, also it seems I need to add support for JSON configs in addition to TOML. I missed that there are a variety of ways to specify datasets.

niceguy4:
Any opportunities with revisiting this implementation? Validation seems to be an important element for verifying training. It appears to work well when used in conjunction with the tensor board loss chart.

Sana-A-E:
What is the state of this? If I pull this branch, will the training work properly and log validation loss properly?
Are you guys still working on getting this merged into the main branch?
Honestly, having a validation loss chart would really help out in picking checkpoints. When training image generation models, you can just output a grid to compare the outputs of all the checkpoints, but you can't do the same with video, which makes comparing checkpoints a chore. Not to mention the time it would take to generate all those videos for comparison. 
With validation loss curve, instead of struggling to compare 20 versions of the checkpoint, it could be cut down to e.g. 4.

PR #88, opened 2025-02-17
Add multi-prompt video generation support from prompt file.

glmcdona:
Adds arguments to `hv_generate_video.py` to allow for specifying a file with a list of prompts to generate videos from:
* `--prompts_file`: Path to text file with prompts (one per line)
* `--increment_seed`: Increment seed for each processed prompt
* `--repeat_count`: Repeat generations of each prompt X times

Also a couple minor documentation tweaks to make it more clear for users.

### DIFF
--- a/hv_generate_video.py
+++ b/hv_generate_video.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from pathlib import Path
 import random
 import sys
+import hashlib
 import os
 import time
 from typing import Optional, Union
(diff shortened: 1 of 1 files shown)


GarbageHaus:
I tested this out and it's really excellent. When generating small videos like 288x480 it seems to increase the generation speed by around 50% as a rough estimate.

If I could make a humble request to have the unique identifier that is inserted in the front of the filename to be appended to the end instead? This makes organization easier since the first part contains a date.

glmcdona:
Note don't merge this yet. I have a couple fixes I need to add first.

PR #137, opened 2025-03-09
Add support for torch.compile and fast fp8 for generation

Sarania:
This PR adds support to hv_generate_video.py for torch.compile and fast fp8 math. Torch.compile improves speed and memory consumption on pretty much any hardware and the compile times of around a minute are well worth it when dealing with 10+ minute generations. FP8 acceleration is supported on RTX 4xxx (and probably 5xxx?) and is a very nice boon to speed with no quality hit. Combined they net me a speed boost of around 30% on my RTX 4070 TI Super, with slightly less memory usage as well. 

The fp8_optimization.py and implementation of that is ported from HunyuanVideoWrapper(with attribution in the comments). The torch.compile implementation is my own and targets only the single and double blocks since compiling other parts of the transformer isn't very beneficial but takes longer.

I tested this in various combinations on my own hardware and it seems to work very well and match or outperform the performance I get in ComfyUI! Hope it's helpful, cheers!

### DIFF
--- a/hunyuan_model/fp8_optimization.py
+++ b/hunyuan_model/fp8_optimization.py
@@ -1,0 +1,39 @@
+#based on ComfyUI's and MinusZoneAI's fp8_linear optimization
+#further borrowed from HunyuanVideoWrapper for Musubi Tuner
+import torch
+import torch.nn as nn
+
+def fp8_linear_forward(cls, original_dtype, input):
+    weight_dtype = cls.weight.dtype
+    if weight_dtype in [torch.float8_e4m3fn, torch.float8_e5m2]:
+        if len(input.shape) == 3:
+            target_dtype = torch.float8_e5m2 if weight_dtype == torch.float8_e4m3fn else torch.float8_e4m3fn
+            inn = input.reshape(-1, input.shape[2]).to(target_dtype)
+            w = cls.weight.t()
+
+            scale = torch.ones((1), device=input.device, dtype=torch.float32)
+            bias = cls.bias.to(original_dtype) if cls.bias is not None else None
+
+            if bias is not None:
+                o = torch._scaled_mm(inn, w, out_dtype=original_dtype, bias=bias, scale_a=scale, scale_b=scale)
+            else:
+                o = torch._scaled_mm(inn, w, out_dtype=original_dtype, scale_a=scale, scale_b=scale)
+
+            if isinstance(o, tuple):
+                o = o[0]
+
+            return o.reshape((-1, input.shape[1], cls.weight.shape[0]))
+        else:
+            return cls.original_forward(input.to(original_dtype))
+    else:
+        return cls.original_forward(input)
+
+def convert_fp8_linear(module, original_dtype, params_to_keep={}):
+    setattr(module, "fp8_matmul_enabled", True)
+   
+    for name, module in module.named_modules():
+        if not any(keyword in name for keyword in params_to_keep):
+            if isinstance(module, nn.Linear):
+                original_forward = module.forward
+                setattr(module, "original_forward", original_forward)
+                setattr(module, "forward", lambda input, m=module: fp8_linear_forward(m, original_dtype, input))
--- a/hv_generate_video.py
+++ b/hv_generate_video.py
@@ -25,6 +25,7 @@ from hunyuan_model.text_encoder import TextEncoder
 from hunyuan_model.text_encoder import PROMPT_TEMPLATE
 from hunyuan_model.vae import load_vae
 from hunyuan_model.models import load_transformer, get_rotary_pos_embed
+from hunyuan_model.fp8_optimization import convert_fp8_linear
 from modules.scheduling_flow_match_discrete import FlowMatchDiscreteScheduler
 from networks import lora
 
@@ -479,6 +480,13 @@ def parse_args():
     parser.add_argument("--no_metadata", action="store_true", help="do not save metadata")
     parser.add_argument("--latent_path", type=str, nargs="*", default=None, help="path to latent for decode. no inference")
     parser.add_argument("--lycoris", action="store_true", help="use lycoris for inference")
+    parser.add_argument("--fp8_fast", action="store_true", help="Enable fast FP8 arthimetic(RTX 4XXX+)")
+    parser.add_argument("--compile", action="store_true", help="Enable torch.compile")
+    parser.add_argument(
+        "--compile_args", nargs=4, metavar=("BACKEND", "MODE", "DYNAMIC", "FULLGRAPH"),
+        default=["inductor", "max-autotune-no-cudagraphs", "False", "False"],
+        help="Torch.compile settings"
+    )
 
     args = parser.parse_args()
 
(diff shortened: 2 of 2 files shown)


Sarania:
Awesome, if it ends up being satisfactory I can add these to wan_generate_video too I just haven't done much wan yet so honestly it wasn't til after I made this PR when I remembered that!

FurkanGozukara:
@Sarania thank you

what is the logic of torch compile how does it work?

Sarania:
> I would rather prioritize Wan2.1, which seems to have more functionality. But it will probably be easier to apply it based on this PR, so I think I can do it after merging this.
> 
> As for HunyuanVideo, I found that they have fp8 optimization code in their official repository. I'm not sure, which is more suitable. https://github.com/Tencent/HunyuanVideo/blob/main/hyvideo/modules/fp8_optimization.py

So my understanding is the convert_fp8_linear in the official repo is for used with the officially released fp8_scaled weights. According to kijai they are more accurate than fp8_e4m3fn weights but the problem is with LoRA compatibility and acceleration:

"Even if this file is .pt it's completely safe and it is loaded with weights_only, the scale map is included with the nodes. To use this model you have to use the fp8_scaled -quantization option in the model loader. The quality of these weights is much closer to the original bf16, downside is that they do not currently support fp8 fast mode, or LoRAs."

Looking at the fp8_optimization.py you linked, it seems that's what it's for. It does at least set fp8_matmul enabled but the actual forward hook is scaling the weights based on the map, nothing to do with accelerating the inference. 

Ultimately LoRA compatibility is likely fundamental priority to maintain here, given the nature of the project, so I suspect the version in the PR is likely superior for this reason.

@FurkanGozukara Torch.compile is a component of pytorch that optimizes the model's code for your hardware by using just in time compilation techniques. Ultimately the end result is you have a small delay before starting inference, but a noticeably faster, more memory efficient inference when it starts. It's just a way to help the model take full advantage of whatever hardware is available. That's the basics, there are also multiple backends like tensorrt, cudagraphs, etc and I did add an arg to specify the settings for it, but I also set sensible defaults which should work for the vast majority of users.

FurkanGozukara:
@Sarania thank you so much so it happens before inference . i assume it is model based not per prompt?

Sarania:
> @Sarania thank you so much so it happens before inference . i assume it is model based not per prompt?

It's complex, it's actually kinda per "shape" in that your res, length, prompt, etc all affect it. That's what the "Dynamic" flag is about - if True it will try to make the compilation as dynamic as possible, to avoid recompiling every inference. But because this is a script that nominally generates once and then completes, that's kinda moot here hence why it's default False. With image generation, the compile time to generation ratio is much more of a concern so you wanna avoid recomp as much as possible(it doesn't make sense to compile for a minute to save 5 seconds every time...). But with video, the gen times are already quite long so it's kind of free, it saves more time than it takes even if you have to compile every time. Here I've also only compiled the single and double blocks because the other parts don't net any real speed boost or mem savings but increase compile time 50%.

FurkanGozukara:
> > @Sarania thank you so much so it happens before inference . i assume it is model based not per prompt?
> 
> It's complex, it's actually kinda per "shape" in that your res, length, prompt, etc all affect it. That's what the "Dynamic" flag is about - if True it will try to make the compilation as dynamic as possible, to avoid recompiling every inference. But because this is a script that nominally generates once and then completes, that's kinda moot here hence why it's default False. With image generation, the compile time to generation ratio is much more of a concern so you wanna avoid recomp as much as possible(it doesn't make sense to compile for a minute to save 5 seconds every time...). But with video, the gen times are already quite long so it's kind of free, it saves more time than it takes even if you have to compile every time. Here I've also only compiled the single and double blocks because the other parts don't net any real speed boost or mem savings but increase compile time 50%.

thank you so much for detailed explanation

Sarania:
> @Sarania Thank you for detailed explanation! In my understanding, this PR is mainly aimed at accelerating inference, while HunyuanVideo's optimization seem to be more comprehensive, so I think the two can be applied independently.
> 
> I created a new PR inspired by HunyuanVideo optimization: #141.
> 
> I will merge this PR separately.

Having both as an option is absolutely possible! Indeed HunyuanVideoWrapper which the code was ported from supports either or! I'm glad you found it helpful! 

I actually was working on a different branch today extending hv_generate_video further with support for changing the LLM arguments (hidden_state_skip layer etc), and also with a facility for previewing the latents during generation(working but not fully integrated nicely, took me a bit to figure out how to get early previews looking good). I'm doing it for my own purposes because I really like the script it's powerful and I may make a frontend to it, plus learning. But I don't know if you'd want me to PR those changes or not because I don't know what your future plans are. I'm absolutely willing, or I can just publish them to a branch on my own fork where you can use whatever bits you need. Just let me know, whatever is helpful!

Sarania:
> > I actually was working on a different branch today extending hv_generate_video further with support for changing the LLM arguments (hidden_state_skip layer etc), and also with a facility for previewing the latents during generation(working but not fully integrated nicely, took me a bit to figure out how to get early previews looking good). I'm doing it for my own purposes because I really like the script it's powerful and I may make a frontend to it, plus learning. But I don't know if you'd want me to PR those changes or not because I don't know what your future plans are. I'm absolutely willing, or I can just publish them to a branch on my own fork where you can use whatever bits you need. Just let me know, whatever is helpful!
> 
> Thank you, that function seems very useful!
> 
> It is true that I want to keep the script's functionality simple. Also, if I merge a function that I don't fully understand, it will be difficult to maintain in the future.
> 
> However, I think it is possible to create some kind of extension feature. I would like to see the code first, so could you please publish it in your own fork?

Sure thing I need to clean it up so I will publish it tomorrow most likely. I also added a simple detection and conversion for diffusers format LoRAs just importing and using from convert_lora.py after detecting the prefix. But yeah I'll publish that tomorrow and you can have a look I got the latent2RGB factors from HunyuanVideoWrapper but the rest of the preview logic I actually sussed out myself. Basically it just takes the current_latents - (original_latents * timesteps[current_step]), normalizes that and then applies the RGB factors via a linear transform to produce a /really/ good preview, like by step 10 of 50 I can tell if I've got a gen I wanna continue with or not and by step 25 you can tell what the finished video will be! I've currently just got it writing out a video latent_preview.mp4 with av, that's the part I'm especially not sure how you'd wanna handle. But I agree it's highly useful and I worked hard on it!

Sarania:
> This PR is merged. Thank you for this!

You're very welcome! Glad it could help!

Sarania:
@kohya-ss Just to let you know, I updated the latent preview code to make it more robust and add support for Wan as well. It's still in the same place: https://github.com/Sarania/musubi-tuner/blob/add-features/blissful_tuner/latent_preview.py but it's a lot more finished now. Cheers.

Sarania:
That definitely sounds like an awesome idea but yeah no rush, please don't interpret any of my posts about this stuff as any kind of pressure or demand! There's so much new stuff going on that I can't keep up with /using/ it let alone implementing it and I also understand that you don't always get to decide what your brain wants to work on sometimes it just fixates on what's most interesting and so that's what we do!

PR #141, opened 2025-03-10
add FP8 optimization support in Wan2.1

kohya-ss:
Add FP8 optimization inspired from https://github.com/Tencent/HunyuanVideo/blob/main/hyvideo/modules/fp8_optimization.py and diffusion-pipe.

Also supports FP16 weights for Wan2.1.

Related: #131 (comment), #137

## Key Features

- Implements FP8 (E4M3) weight quantization for Linear layers
- Maintains computational precision by dequantizing to original precision (FP16/BF16/FP32) during forward pass
- Includes DiT blocks, and excludes modules require higher precision.

## Implementation Details

1. Quantizes weights to FP8 format with appropriate scaling
2. Stores both quantized weights and scale factors in model state dict
3. Applies monkey patching to Linear layers for transparent dequantization during computation


scottmudge:
Would it be a bad idea to merge this PR into my local repo and resume training a 14B I2V model I've already got 10 epochs trained on? Or will the improved precision only improve additional training?

Sarania:
I actually ported this excellent work to WanVideoWrapper:

https://github.com/kijai/ComfyUI-WanVideoWrapper/pull/266

Sarania:
> > I actually ported this excellent work to WanVideoWrapper:
> 
> That's great! I'm glad to see this is useful for others too.

I'm quite fond of it yeah, it gives me the closest results to unquantized I've gotten! And I actually ported it to HunyuanVideoWrapper today too, so, we've come full circle! 🤗

Sarania:
Sorry to keep pinging this thread over and over but today I ported the fp8 scaling code to my personal AI agent platform and... wow. It is SO good. It's such a robust implementation and I love it. For LLM, I use a finetune of Gemma2 9B that I further customized for my own needs and it's just a bit big to run unquantized in 16GB I mean you can with offloading but it's not efficient. BNB exists but doesn't like torch.compile and is picky about various configurations, but this code is highly effective, easy to approach and understand and yet remains highly performant. It's beautiful! 🥰

PR #144, opened 2025-03-11
Read t5 tokenizer from weight_path

TinyTigerPan:
Read t5 tokenizer from weight_path, do not need download.


TinyTigerPan:
I also find a bug on `wan_generate_video.py`
The type float will lead `len(args.lora_multiplier)` raise error in Line 251.

TinyTigerPan:
Thanks for your reply. I got it.

PR #153, opened 2025-03-13
feat: refactor wan2.1 generation script, add compile option

kohya-ss:
- Refactor wan_generate_video.py, image2video.py and text2video.py to a single script.
- Add `--compile` option for wan_generate_video.py


Sarania:
Nice, it's worth noting that the fp8 fast that I recently PR for HyVideo doesn't seem to play nice with Wan, at least as we currently understand it. You may have encountered that yourself. There is an fp16 accumulation option in Pytorch nightlies past mid February that's being called "fp16_fast" in Wan wrapper and noticeably accelerates inference with Wan. But the Pytorch nightly has broken torch.compile for Hyvideo, at least for me, still works for Wan though. Welp, another venv enters the fray heh. Anyway you can see that implemented into hv_generate_video here:

https://github.com/Sarania/musubi-tuner/commit/e97d1416f22c8e7feed1524a53593b68a6e1e660

Although personally I don't think it's useful for HyVideo at least until the compile gets fixed and even then only when not using fp8_fast, I prolly won't keep it in my fork, but it might be useful for Wan and seems fairly easy to implement. Always up to you, just sharing what I got. I'm training Wan tonight with the new fp8_scaled option to see how that goes!

Sarania:
I tried the fp8_scaled option as it is in the main branch last night and got REALLY good results. It was my first time training Wan at all but I used similar settings to what I had been using with Hunyuan except lower res because VRAM and I saw much quicker, tighter convergence than I've seen when training video before. I don't know if this is typical of Wan because CFG, or how much the fp8_scaled helped but it's excellent!

PR #161, opened 2025-03-16
add torch.no_grad() for block swap

kohya-ss:
This may solve a potential issue and prevent loss=NaN in long training. Further testing is needed to see if existing training works as well as before.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -39,6 +39,12 @@ Wan2.1については、[Wan2.1のドキュメント](./docs/wan.md)も参照し
 
 ### 最近の更新
 
+- 2025/03/18
+    - SageAttentionのインストール方法が最新に更新されました（ソースコードの編集が不要となりました）。PR [#165](https://github.com/kohya-ss/musubi-tuner/pull/165) fai-9氏に感謝いたします。
+
+- 2025/03/17
+    - Wan2.1の学習で、float8_e4m3fnの重みを使用した場合に、LoRAの重みも同形式になり、正しく保存されない不具合を集成しました。
+
 - 2025/03/16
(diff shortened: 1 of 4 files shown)


FurkanGozukara:
> This fix appears to enable multi-GPU training of DDP, but further testing is required.

with flux we were never able to do multi gpu training with block swap

so it is possible?

FurkanGozukara:
> > so it is possible?
> 
> I think so. I will add a new branch to support this on sd-scripts repo today.

awesome

xzuyn:
I've been using this on a single amd gpu on ubuntu for a bit, training qwen image. Haven't noticed any problems.

modules/custom_offloading_utils.py
```diff
@@ -56,19 +56,20 @@ def swap_weight_devices_cuda(device: torch.device, layer_to_cpu: nn.Module, laye
...
-            cuda_data_view.record_stream(stream)
-            module_to_cpu.weight.data = cuda_data_view.data.to("cpu", non_blocking=True)
-
-        stream.synchronize()
-
-        # cpu to cuda
-        for module_to_cpu, module_to_cuda, cuda_data_view, cpu_data_view in weight_swap_jobs:
-            cuda_data_view.copy_(module_to_cuda.weight.data, non_blocking=True)
-            module_to_cuda.weight.data = cuda_data_view
+    with torch.no_grad():
```
hameerabbasi, line 59:
```suggestion
    with torch.inference_mode():
```

See https://discuss.pytorch.org/t/pytorch-torch-no-grad-vs-torch-inference-mode/134099/2

PR #181, opened 2025-03-22
feat: Add a new extraction method 'full' for the video dataset

kohya-ss:
This will be useful when each video represents a single complete motion, and has different length.

Also add a feature to trim `target_frames` to 4*N+1 automatically.


FurkanGozukara:
great update. anywhere it saves these extracted frames so users can see?

FurkanGozukara:
> > great update. anywhere it saves these extracted frames so users can see?
> 
> Unfortunately each fragments are saved as latent. You can run `cache_latents.py` or `wan_cache_latents.py` with `--debug_mode image` or `--debug_mode console` option to verify the first and end frame of each fragment.

any chance you can add this option this is extremely necessary. otherwise people has to use imagination :D 

and debug mode is not really that helpful. i am assuming adding this would be a trivial task

FurkanGozukara:
> I added `--debug_mode video` in #187. I thought it would probably be better to check before doing the actual caching.

thank you so much great addition

PR #186, opened 2025-03-23
Skip layer guidance for Wan2.1

kohya-ss:
`original` (same as SD3.5) and `uncond` (same as Wan2GP) are implemented.


maybleMyers:
Sure I will try to with the updated wan_generate_video.py from your recent pr.

PR #213, opened 2025-04-06
Extension mechanism for video generation script

kohya-ss:
Highly experimental and may not be merged.

### DIFF
--- a/utils/built_in_extensions.py
+++ b/utils/built_in_extensions.py
@@ -1,0 +1,25 @@
+from extensions import Extension
+
+
+def load_builtin_extensions(ext_name: str, ext_config: dict):
+    """Load built-in extensions."""
+
+    built_in_extensions = {
+        "no_op_extension": Extension,
+        "cfg_skip": CfgSkipExtension,
+    }
+    if ext_name in built_in_extensions:
+        extension_class = built_in_extensions[ext_name]
+        return extension_class(ext_config)
+    else:
+        raise ValueError(f"Extension {ext_name} not found in built-in extensions.")
+
+
+class CfgSkipExtension(Extension):
+    """A built-in extension that skips the configuration."""
+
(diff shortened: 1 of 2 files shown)


Auswolf-debug:
> honestly I keep trying to use ComfyUI because of the suite of tools available there but keep coming back to Musubi because I simply get the best results with it by far. 

I was thinking the same thing....

Sarania:
> > honestly I keep trying to use ComfyUI because of the suite of tools available there but keep coming back to Musubi because I simply get the best results with it by far.
> 
> I was thinking the same thing....
> 
> Screenshot from 2025-04-08 08-05-40
> 
> Screenshot from 2025-04-08 08-17-03

YO THAT'S SICK. I really hope you share that!

Auswolf-debug:
> This is a different approach, but it's great. If you need any changes to the script for this, please let me know and I'll do my best to update.

All good, shouldn't need any changes.

> YO THAT'S SICK. I really hope you share that!

Hopefully have it in comfyui manager within a month. Need to make the inference workflows more user friendly.

Sarania:
I'm working on converting my latent previewer into a class that will more easily be able to be made into an extension. Now with tiny autoencoder powers! I'll endeavour to make it the first extension when we get things sorted.

@Auswolf-debug that's awesome!

Auswolf-debug:
Well I've run into a problem and cant find a fix :(

It only effects Wan2.1 v2v using the fun control models,  also seems to be a mismatch between tasks and actual available models. From tasks t2v-1.3B-FC, t2v-14B-FC and i2v-14B-FC but only 2 models available from the linked website.?

Anyway using either 1.3b or the 1.4b version, results are the same.
Screenshot from 2025-04-13 12-00-43

Not sure if anyone has seen this before, but  starting to think its a environment problem.

PR #215, opened 2025-04-07
Add dynamo for torch.compile

sdbds:
Considering that video training often has a single resolution, we can directly use torch.compile to reduce memory usage and increase speed.


pyproject.toml
```diff
@@ -5,7 +5,7 @@ description = "Musubi Tuner by kohya_ss"
 readme = "README.md"
 requires-python = ">=3.10, <3.11"
 dependencies = [
-    "accelerate>=1.0.0",
+    "accelerate>=1.2.1",
```
kohya-ss: Is it okay to update requirements.txt as well?

sdbds: I think the requirements should also be upgraded, I modified it locally and forgot to commit it.

sdbds:
I just checked the accelerate logs, they fixed a memory leak issue in version 1.4.
We should upgrade to at least version 1.4 or above.

Sarania:
Very nice I was hoping we could leverage this at some point, it's a shame dynamic is not doing well but still. One thing I do wonder about is the way you are specifying the args. Nothing wrong with it per se it's just different than how I did it on the inference side of things:

```
    parser.add_argument(
        "--compile_args",
        nargs=4,
        metavar=("BACKEND", "MODE", "DYNAMIC", "FULLGRAPH"),
        default=["inductor", "max-autotune-no-cudagraphs", "False", "False"],
        help="Torch.compile settings",
    )
```
I would suggest we use the same format for both whichever format that may be. I chose to use a single arg just to avoid adding 6 args for one feature but if you wanna go the more verbose explicit route, that's understandable too. I just think whichever way, we should make them the same!

sdbds:
> Very nice I was hoping we could leverage this at some point, it's a shame dynamic is not doing well but still. One thing I do wonder about is the way you are specifying the args. Nothing wrong with it per se it's just different than how I did it on the inference side of things:
> 
> ```
>     parser.add_argument(
>         "--compile_args",
>         nargs=4,
>         metavar=("BACKEND", "MODE", "DYNAMIC", "FULLGRAPH"),
>         default=["inductor", "max-autotune-no-cudagraphs", "False", "False"],
>         help="Torch.compile settings",
>     )
>     ...
>     ...
>     compile_backend, compile_mode, compile_dynamic, compile_fullgraph = args.compile_args
> ```
> 
> I would suggest we use the same format for both whichever format that may be. I chose to use a single arg just to avoid adding 6 args for one feature but if you wanna go the more verbose explicit route, that's understandable too. I just think whichever way, we should make them the same!

I think four parameters are better because it's more convenient to handle errors, and you can quickly find out on which option the potential BUG occurred.
And for ordinary users, they may not need advanced options.

sdbds:
> Oh can I ask @sdbds why no "max-autotune-no-cudagraphs" option for the mode? It does the same thing as "max-autotune" but doesn't use cudagraphs which can take up additional VRAM. The document looks good to me the one addition I might suggest is sdbds mentioned
> 
> "but after testing, I found that dynamic compilation has many problems."
> 
> which might be worth mentioning along with that setting?

Because here, the plugin method of accelerate is used to pass dynamo parameters, rather than directly through torch settings.
The training mainly uses accelerate to manage model states, and currently, there is no option of max-autotune-no-cudagraphs in its documentation.
https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/dataclasses.py#L969

Sarania:
Understood, thank you for explaining!

PR #312, opened 2025-05-27
Fix big bug when just match one of hw backet

sdbds:
If only one wh is matched, the other wh will not be matched.

```
crop_left = (image_width - bucket_width) // 2
crop_top = (image_height - bucket_height) 2
```

For example 1280,880 with 1280,896
crop_left = (1280 - 1280) = 0
crop_top = (880 - 896) = -8

### DIFF
--- a/dataset/image_video_dataset.py
+++ b/dataset/image_video_dataset.py
@@ -124,23 +124,21 @@ def resize_image_to_bucket(image: Union[Image.Image, np.ndarray], bucket_reso: t
         return np.array(image) if is_pil_image else image
 
     bucket_width, bucket_height = bucket_reso
-    if bucket_width == image_width or bucket_height == image_height:
-        image = np.array(image) if is_pil_image else image
-    else:
-        # resize the image to the bucket resolution to match the short side
-        scale_width = bucket_width / image_width
-        scale_height = bucket_height / image_height
-        scale = max(scale_width, scale_height)
-        image_width = int(image_width * scale + 0.5)
-        image_height = int(image_height * scale + 0.5)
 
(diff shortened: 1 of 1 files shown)


dataset/image_video_dataset.py
```diff
@@ -124,23 +124,21 @@ def resize_image_to_bucket(image: Union[Image.Image, np.ndarray], bucket_reso: t
         return np.array(image) if is_pil_image else image
 
     bucket_width, bucket_height = bucket_reso
-    if bucket_width == image_width or bucket_height == image_height:
```
kohya-ss, line 127: Perhaps changing `or` to `and` here would solve the issue?

sdbds: > Perhaps changing `or` to `and` here would solve the issue?也许将这里的 `or` 改为 `and` 可以解决这个问题？

That's so, but I think this is repetitive with the previous two lines of code...
https://github.com/kohya-ss/musubi-tuner/blob/8684dc58ace18f40d9a3f78d4795bcbf278da0dc/dataset/image_video_dataset.py#L123-L124

kohya-ss: Oh, sorry, I got it!

arledesma:
This affected both latent caching and generation, right? 

No resizing would occur when one side of the image matched?

sdbds:
> This affected both latent caching and generation, right?
> 
> No resizing would occur when one side of the image matched?

Yes, that's how it was originally handled. Now, it only skips scaling when they are exactly the same, otherwise, the cropping dimension of the latent will have issues. 
like crop_left,crop_top

PR #327, opened 2025-06-02
fix: FramePack one frame inference crashes without no_2x or 4x

### LINKED ISSUE #326: 1frame inference without control image, no_2x, no_4x option broken
The single frame inference without controlimage and execution without the option to skip clean latent are broken. I think the implementation needs to be changed as shown below, but I have not tested the effect on controlnet inference, etc. For reference.
```
    if control_latents is None or len(control_latents) == 0:
        # No control images provided, use start_latent from --image_path instead
        image_index = 0  # Use the first (primary) image
        if image_index in context_img and "start_latent" in context_img[image_index]:
            start_latent = context_img[image_index]["start_latent"]
            logger.info(f"No control images provided for one frame inference. Using start_latent from --image_path: {context_img[image_index].get('image_path', 'unknown')}")
            control_latents = [start_latent]
        else:
            logger.info(f"No control images or start_latent provided for one frame inference. Use zero latents for control images.")
            control_latents = [torch.zeros(1, 16, 1, height // 8, width // 8, dtype=torch.float32)]
```

```
    if "no_2x" in one_frame_inference:
        clean_latents_2x = None
        clean_latent_2x_indices = None
        logger.info(f"No clean_latents_2x")
    else:
        clean_latents_2x = torch.zeros((1, 16, 2, height // 8, width // 8), dtype=torch.float32)
        index = 1 + latent_window_size + 1
        clean_latent_2x_indices = torch.arange(index, index + 2).unsqueeze(0)  # Shape (1, 2) to match latent_indices

    if "no_4x" in one_frame_inference:
        clean_latents_4x = None
        clean_latent_4x_indices = None
        logger.info(f"No clean_latents_4x")
    else:
        clean_latents_4x = torch.zeros((1, 16, 16, height // 8, width // 8), dtype=torch.float32)
        index = 1 + latent_window_size + 1 + 2
        clean_latent_4x_indices = torch.arange(index, index + 16).unsqueeze(0)  # Shape (1, 16) to match latent_indices

```

kohya-ss:
closes #326

### DIFF
--- a/fpack_generate_video.py
+++ b/fpack_generate_video.py
@@ -1348,15 +1348,16 @@ def generate_with_one_frame_inference(
     else:
         clean_latents_2x = torch.zeros((1, 16, 2, height // 8, width // 8), dtype=torch.float32)
         index = 1 + latent_window_size + 1
-        clean_latent_2x_indices = torch.arange(index, index + 2)  #  2
+        clean_latent_2x_indices = torch.arange(index, index + 2).unsqueeze(0)  #  2
 
     if "no_4x" in one_frame_inference:
         clean_latents_4x = None
         clean_latent_4x_indices = None
         logger.info(f"No clean_latents_4x")
     else:
+        clean_latents_4x = torch.zeros((1, 16, 16, height // 8, width // 8), dtype=torch.float32)
         index = 1 + latent_window_size + 1 + 2
-        clean_latent_4x_indices = torch.arange(index, index + 16)  #  16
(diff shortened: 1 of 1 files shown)


sdbds:
I think the training part also needs to be modified for this, otherwise the sample will have issues.

Also, I noticed that after the recent update, for some reason, the input image is completely different from the result...

kohya-ss:
Thank you for reporting! The sample generation is fixed in #339.

The generation script now requires control_image for one frame generation as pre clean latents. It seems I forgot to update the documentation about this, sorry about that. I will update it sooner.

Issue #329, opened 2025-06-04
Suggestion for easy improvement of interactive modes!


Sarania:
I've been adding support for prompt_toolkit to the interactive mode of fpack and wan in Blissful Tuner. Unlike input, it allows editing the line as you write it and recalling previous prompts from this session and the implementation is quite simple it's basically a drop in replacement for input:

```
diff --git a/wan_generate_video.py b/wan_generate_video.py
index c5b43ad..1c27747 100644
--- a/wan_generate_video.py
+++ b/wan_generate_video.py
@@ -20,7 +20,7 @@ import cv2
 import numpy as np
 import torchvision.transforms.functional as TF
 from tqdm import tqdm
-
+from prompt_toolkit import PromptSession
 from networks import lora_wan
 from utils.safetensors_utils import mem_eff_save_file
 from wan.configs import WAN_CONFIGS, SUPPORTED_SIZES
@@ -1664,12 +1664,13 @@ def process_interactive(args: argparse.Namespace) -> None:
     model = None
     clip = None

+    session = PromptSession()
     print("Interactive mode. Enter prompts (Ctrl+D to exit):")

     try:
         while True:
             try:
-                line = input("> ")
+                line = session.prompt("> ")
                 if not line.strip():
                     continue
```

It does require prompt_toolkit installed in the venv but that's cross platform, fully python so it's not a big ask. As always it's up to you but this is a big QOL improvement for little effort that I feel massively improves the utility of interactive mode 🙂 I was completely unaware of this prompt_toolkit until AI told me about it so I thought to share! Cheers!

kohya-ss:
Thank you for sharing!

Hmm... I tried prompt-toolkit, but I didn't notice any significant improvement. I can refer to previous prompts with the up and down arrows with `input("> ")`. In fact, I find it easier to use because it remains previous prompts as the history (when I edit "prompt A" to "prompt B", the up arrow can recall both "prompt A" and "prompt B").

Maybe I'm not using prompt-toolkit features properly. Can you tell me a bit more about how to use this?

P.S I made a draft PR for this:  #330

Sarania:
[...]

PR #330, opened 2025-06-04
Enhance interactive input handling with prompt_toolkit support

kohya-ss:
ref #329


Sarania:
I mentioned it in the post but yeah this looks great! Since Windows has the base line editing features they don't have to worry about it but Linux users can become aware of prompt-toolkit because having the line editing features is the difference between interactive being useful versus annoying!

PR #361, opened 2025-06-21
fix: Correctly handle four-channel images in encode_and_save_batch(_o…

zhristophe:
## Problem Description
The current implementation fails to handle four-channel (RGBA) images correctly during batch processing in the `encode_and_save_batch(_one_frame)` function.

## Solution
```python
if image.shape[-1] == 4:
    image = image[..., :3]
```


kohya-ss:
Thank you for this! When we made another function (one-frame training) compatible with RGBA, we forgot to fix this.

PR #402, opened 2025-08-01
Fix bugs about flux kontext part

sdbds:
1、change order for right position.
2、transform str to torch.device avoid error


kohya-ss:
Thank you! I forgot to test block swap.

PR #407, opened 2025-08-03
Add new SNR function for better training

sdbds:
logsnr from paper:https://arxiv.org/abs/2411.14793v3
default log_mean=-6 and log_std=2
can better learning style.

I tried mixing the proportions and increased the learning of the low-frequency part, the ratio is about 2:7:1.
image

After my testing, it seems the details are better.


kohya-ss:
Thank you, this is quite interesting! I will merge this sooner.

PR #428, opened 2025-08-11
Add qwen_shift and qinglong_qwen

sdbds:
The feature for freely controlling the ratio of lows, mids, and highs to freely change qinglong's ratio will be considered for addition later.

### DIFF
--- a/src/musubi_tuner/hv_train_network.py
+++ b/src/musubi_tuner/hv_train_network.py
@@ -765,6 +765,7 @@ class NetworkTrainer:
             or args.timestep_sampling == "sigmoid"
             or args.timestep_sampling == "shift"
             or args.timestep_sampling == "flux_shift"
+            or args.timestep_sampling == "qwen_image_shift"
             or args.timestep_sampling == "logsnr"
             or args.timestep_sampling == "qinglong"
         ):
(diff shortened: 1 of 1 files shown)


FurkanGozukara:
so many options

choices=["sigma", "uniform", "sigmoid", "shift", "flux_shift", "qwen_shift", "logsnr", "qinglong_flux", "qinglong_qwen"],

according to each the value of shift would change too right?

which values you would recommend @sdbds for Qwen Image

sdbds:
> so many options
> 
> choices=["sigma", "uniform", "sigmoid", "shift", "flux_shift", "qwen_shift", "logsnr", "qinglong_flux", "qinglong_qwen"],
> 
> according to each the value of shift would change too right?
> 
> which values you would recommend @sdbds for Qwen Image

qwen_shift is the original option, while qinglong_qwen is an experimental option I modified. You can choose one of these two

FurkanGozukara:
@sdbds thank you. qwen_shift recommended value is 2.2? original?

sdbds:
> @sdbds thank you. qwen_shift recommended value is 2.2? original?

This is an automatic resolution option that you don't need to set specifically. When the standard resolution is 1328X1328, it is equivalent to manually setting 2.2.

kohya-ss:
Thank you for this!

PR #493, opened 2025-08-24
feat: Wan use original dtype in AttentionBlock to reduce memory usage

kohya-ss:
Depending on the frame size and number of frames, the peak memory requirement seems to be reduced by several GB during training and about a quarter of that during inference.


kohya-ss:
Technical details:

In some parts of AttentionBlock, tensors were of type float32 during calculations, but were explicitly cast to type float16. These values ​​were originally implicitly cast to float16 in the following operation, so there should be almost no impact on the calculation (training/inference) results.

Enyakk:
Excellent update. Early training results in much larger batch sizes while retaining quality.

PR #506, opened 2025-08-27
trying to implement adding captions to model metadata

SamTyurenkov:
im just about to test it, will post if it works or not additionally. and update the branch if needed

### DIFF
--- a/src/musubi_tuner/hv_train_network.py
+++ b/src/musubi_tuner/hv_train_network.py
@@ -541,7 +541,7 @@ class NetworkTrainer:
     def is_schedulefree_optimizer(self, optimizer: torch.optim.Optimizer, args: argparse.Namespace) -> bool:
         return args.optimizer_type.lower().endswith("schedulefree".lower())  # or args.optimizer_schedulefree_wrapper
 
-    def get_dummy_scheduler(optimizer: torch.optim.Optimizer) -> Any:
+    def get_dummy_scheduler(self, optimizer: torch.optim.Optimizer) -> Any:
         # dummy scheduler for schedulefree optimizer. supports only empty step(), get_last_lr() and optimizers.
         # this scheduler is used for logging only.
         # this isn't be wrapped by accelerator because of this class is not a subclass of torch.optim.lr_scheduler._LRScheduler
@@ -2066,7 +2066,7 @@ class NetworkTrainer:
 
         # TODO refactor metadata creation and move to util
         metadata = {
-            "ss_" "ss_session_id": session_id,  # random integer indicating which group of epochs the model came from
+            "ss_session_id": session_id,  # random integer indicating which group of epochs the model came from
             "ss_training_started_at": training_started_at,  # unix timestamp
             "ss_output_name": args.output_name,
             "ss_learning_rate": args.learning_rate,
@@ -2267,6 +2267,8 @@ class NetworkTrainer:
         logger.info(f"DiT dtype: {transformer.dtype}, device: {transformer.device}")
 
         clean_memory_on_device(accelerator.device)
+
+        optimizer_train_fn()  # Set training mode
 
         for epoch in range(epoch_to_start, num_train_epochs):
             accelerator.print(f"\nepoch {epoch+1}/{num_train_epochs}")
--- a/.ai/codex.prompt.md
+++ b/.ai/codex.prompt.md
@@ -1,0 +1,27 @@
+## About This File
+
+This file provides guidance to Codex CLI when working with code in this repository.
+
+## Guidelines
+
+### Coding Style & Naming Conventions
+- Style: PEP 8, 4‑space indentation, limit lines to ~120 chars.
+- Naming: snake_case for files/functions (`*_train_network.py`, `*_generate_*`), PascalCase for classes.
+- Types/Docs: Prefer type hints for public APIs and short docstrings describing args/returns.
+- Formatting: No formatter configured; keep diffs small and consistent with surrounding code.
+
+### Testing Guidelines
+- Current state: No formal test suite.
+- If adding tests, use `pytest`, place under `tests/` mirroring `src/musubi_tuner/` and name files `test_*.py`.
+- Run (uv): `uv run pytest -q`. Run (pip): `pytest -q`.
+- Prefer small, deterministic unit tests around data utilities and argument parsing.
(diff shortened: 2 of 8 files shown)


SamTyurenkov:
Well, it seems to be working:

```
  "ss_datasets": [
    {
      "batch_size_per_device": 1,
      "bucket_no_upscale": false,
      "caption_count": 2,
      "caption_extension": null,
      "enable_bucket": true,
      "frame_extraction": "full",
      "frame_sample": 1,
      "frame_stride": 1,
      "has_control": false,
      "max_frames": 81,
      "num_repeats": 2,
      "resolution": [
        1024,
        512
      ],
      "sample_captions": [
        "My caption 1.",
        "My caption 2."
      ],
      "source_fps": 16,
      "target_frames": [
        1
      ],
      "video_jsonl_file": "dataset_16fps_1024x512.jsonl"
    },
```

.cursor/rules/project-rule.mdc
kohya-ss, line 1: Please add `.cursor` to `.gitignore`.

src/musubi_tuner/hv_train_network.py
```diff
@@ -1979,12 +2095,28 @@ def load_model_hook(models, input_dir):
...
             dataset_metadata = dataset.get_metadata()
+            
+            # Extract captions from JSONL datasets for trigger word analysis
+            if hasattr(dataset, 'datasource'):
+                dataset_captions = self._extract_captions_from_dataset(dataset.datasource)
+                all_captions.extend(dataset_captions)
+                if dataset_captions:
+                    dataset_metadata["caption_count"] = len(dataset_captions)
+                    dataset_metadata["sample_captions"] = dataset_captions[:3]  # Save first 3 as examples
+
```
kohya-ss: Could you please set the caption information to `tag_frequency` in the same way as `sd-scripts`? This will allow us to view the captions in the metadata viewer.

I think we can include all captions, and it would be appropriate to treat each caption as a tag and count the number of occurrences of the same caption. However, this may result in `tag_frequency` becoming too large, so I recommend keeping the most frequent entries to around 100KB.

I believe this will maintain compatibility and provide convenience to users.

src/musubi_tuner/hv_train_network.py
kohya-ss, line 1: I'm sorry that this is a specification that is not in the documentation, but for training, image files, caption files, and JSONL files are not necessary, and training will work using only the cached files. So we cannot access caption files or JSON files, but can get captions from Text Encoder cache files. See: 

https://github.com/kohya-ss/musubi-tuner/blob/fec404cb9564853e6abf4882c781a4916e9dfdcf/src/musubi_tuner/dataset/image_video_dataset.py#L401-L434

Also, since accessing all cache files to extract metadata can be time-consuming for large datasets, and users may not want captions stored as metadata, it is better to enable this feature only when a certain option is specified.

It would also be appropriate to aggregate captions on the DataSet side, since the dataset is also used for fine tuning. Since ItemInfo is held by BucketBatchManager, it would be a good idea to delegate the process to that class from get_metadata.

kohya-ss: `tag_frequency` counting is implemented in here in sd-scripts repo: https://github.com/kohya-ss/sd-scripts/blob/1470cb8508a34637aea8dcba7afb286b7caf961f/library/train_util.py#L709-L718

.cursor/rules/project-rule.mdc
SamTyurenkov: @kohya-ss fixed

src/musubi_tuner/hv_train_network.py
SamTyurenkov: @kohya-ss i implemented it in different method, see in this commit https://github.com/kohya-ss/musubi-tuner/pull/506/commits/d2631e838e320b32c24f6fd54f60522536aa03ad

will need to test, to see if it works

src/musubi_tuner/hv_train_network.py
SamTyurenkov: @kohya-ss i don't think fetching captions from cache should be implemented, even if we use text encoder cache for training and don't text encode with every run again - files are still present on disk probably, so we can just access them directly, even if text encoder is cached already. 

Even if you want it, it might be a different PR, as it probably won't break the process to use it as it is now.

src/musubi_tuner/hv_train_network.py
kohya-ss: Thank you, this implementation seems reasonable, but please understand that it may change after the merge.

src/musubi_tuner/hv_train_network.py
kohya-ss: Accessing the text encoder cache is safe because training is not possible without the text encoder cache, and training is possible without captions.
In fact, for example, with cloud training, training is possible simply by uploading the cache files, and it is assumed that there will be no caption files.

src/musubi_tuner/hv_train_network.py
SamTyurenkov: > Thank you, this implementation seems reasonable, but please understand that it may change after the merge.

@kohya-ss it's ok, I also use lora metadataviewer, but i modified it a bit for use on server, its packed in flask app and works with runpod workspace.

image

so probably as long as it will work with metadataviewer or just appear in metadata json - is ok by me i think

src/musubi_tuner/hv_train_network.py
SamTyurenkov: > Accessing the text encoder cache is safe because training is not possible without the text encoder cache, and training is possible without captions. In fact, for example, with cloud training, training is possible simply by uploading the cache files, and it is assumed that there will be no caption files.

why would you upload cache files instead of caption files? it seems like unrealistic behavior, if you train on server, you prepare cache on server. not really reasonable to split work into two machines - prepare cache on one machine and train on other machine.

Also, often you might want to change caption or add to/remove from dataset - and thus creating new cache is a must.
I don't even have text encoder locally, but even if i had I wouldn't split the process.

kohya-ss: Captioning is a relatively light task, so it's practical to do it on a local GPU, and many users would prefer to reduce the time spent on cloud servers.

arledesma:
sweet! I had this same enhancement in my mental backlog. 👏🏽

PR #513, opened 2025-08-29
Add REX learning rate scheduler

xzuyn:
This adds the REX learning rate scheduler from [**REX: Revisiting Budgeted Training with an Improved Schedule**](https://arxiv.org/abs/2107.04197). The implementation was modified from [IvanVassi/REX_LR](https://github.com/IvanVassi/REX_LR).

Use `--lr_scheduler rex` to choose it. Supports `--lr_scheduler_min_lr_ratio` and `--lr_warmup_steps`.

`--lr_scheduler_min_lr_ratio 0.01` is used by default, as without it the first step (if using warmup) and last step will use 0 learning rate.

Example with `--lr_scheduler rex --learning_rate 1e-4 --lr_scheduler_min_lr_ratio 0.01 --lr_warmup_steps 10 --max_train_steps 100`, comparing the paper's values (`0.5+0.5`) and the values IvanVassi's implementation used (`0.1+0.9`). This PR is using `0.1+0.9`.
rex-comparison

### DIFF
--- a/src/musubi_tuner/hv_train_network.py
+++ b/src/musubi_tuner/hv_train_network.py
@@ -613,6 +613,85 @@ class NetworkTrainer:
             # logger.info(f"adafactor scheduler init lr {initial_lr}")
             return wrap_check_needless_num_warmup_steps(transformers.optimization.AdafactorSchedule(optimizer, initial_lr))
 
+        if name.lower() == "rex":
+            class RexLR(torch.optim.lr_scheduler.LRScheduler):
+                """
+                Reflected Exponential (REX) learning rate scheduler (https://arxiv.org/abs/2107.04197)
+                Modified from: https://github.com/IvanVassi/REX_LR
+
+                Args:
+                    optimizer (torch.optim.Optimizer): The optimizer to schedule the learning rate for
+                    max_lr (float): The maximum learning rate
+                    min_lr (float): The minimum learning rate
+                    num_steps (int): The total number of training steps
+                    num_warmup_steps (int): The number of warmup steps
+                    last_epoch (int): The index of the last step
+                """
+
+                def __init__(self, optimizer, max_lr, min_lr=0.0, num_steps=0, num_warmup_steps=0, last_epoch=-1):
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for this PR! Could you please tell me the difference when using the polynomial scheduler with power<1.0?

It seems to draw a curve very similar to REX.

Figure_1

xzuyn:
They are similar, but not the same. REX lets you set the `min_lr` which is respected for warm-up first step and also the last training step. Polynomial starts and ends on two different values with no option for a `min_lr`, and does a sharp drop on the last step.

REX args: `--lr_scheduler rex --learning_rate 1e-4 --lr_scheduler_min_lr_ratio 0 --lr_warmup_steps 10 --max_train_steps 100`
Polynomial args: `--lr_scheduler polynomial --learning_rate 1e-4 --lr_scheduler_power [0.25, 0.5, 0.75] --lr_warmup_steps 10 --max_train_steps 100`
Screenshot 2025-09-03 at 02-53-36 Qwen-Image-musubi Workspace – Weights   Biases

kohya-ss:
Thanks for the clarification. Honestly, I don't think there's much difference...

I wonder if it would be possible to make the two values ​​related to rex_factor into parameters, so that users could choose between the paper implementation, IvanVassi's, or something else. That might be a useful alternative to Polynomial.

xzuyn:
You can now use `--rex_alpha` and `--rex_beta` to adjust the values. It uses IvanVassi's values by default. Let me know if you would rather it use the paper's values, or use different variable names.

kohya-ss:
Thank you! I will review and merge this sooner!

kohya-ss:
Thank you again for this.

It is not desirable to add parameters that depend on a specific scheduler to the training script, so after merging, I would like to change `rex_alpha` and `rex_beta` to be obtained from `--lr_scheduler_args`. I will also move the class to a separate module. Thank you for your understanding.

xzuyn:
Oh I didn't even see that arg. That would have been better to use.

PR #553, opened 2025-09-12
from_file with qwen_image_generate_image is broken

nmfisher:
Using `qwen_image_generate_image` with the `from_file` arg will throw an error. The script doesn't properly handle the result returned by `prepare_text_inputs`.

I've opened this PR as a placeholder while I test the fix. This version works for non-edit mode, but I'll be trying edit mode later today.


kohya-ss:
Thank you for this! I was also working on this issue. After merging this, I will fix other issues sooner.

PR #575, opened 2025-09-18
feat: FP8 quantization with block-wise scaling

kohya-ss:
Default to block-wise scaling fp8 quantization instead of the current per-tensor scaling quantization.

Discussion: #564.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -64,6 +64,12 @@
 
 GitHub Discussionsを有効にしました。コミュニティのQ&A、知識共有、技術情報の交換などにご利用ください。バグ報告や機能リクエストにはIssuesを、質問や経験の共有にはDiscussionsをご利用ください。[Discussionはこちら](https://github.com/kohya-ss/musubi-tuner/discussions)
 
+- 2025/09/23
+    - `--fp8_scaled`オプションを指定した時の量子化方法を、per-tensorからblock-wise scalingに変更しました。[PR #575](https://github.com/kohya-ss/musubi-tuner/pull/575) [Discussion #564](https://github.com/kohya-ss/musubi-tuner/discussions/564)も参照してください。
+        - これによりFP8量子化の精度が向上し、各モデル（HunyuanVideoを除く）学習の安定、推論精度の向上が期待できます。学習、推論速度はわずかに低下します。
+        - Qwen-ImageのLoRA学習では、量子化対象モジュールの見直しにより、学習に必要なVRAMが5GB程度削減されます。
+        - 詳細は[高度な設定のドキュメント](./docs/advanced_config.md#fp8-weight-optimization-for-models--モデルの重みのfp8への最適化)を参照してください。
+
 - 2025/09/22
     - FramePackのVAEについて強制的にtilingが有効になっていた不具合を修正しました。`--vae_tiling`オプションを指定するか、`--vae_spatial_tile_sample_min_size`を指定することでtilingが有効になります。[PR #583](https://github.com/kohya-ss/musubi-tuner/pull/583)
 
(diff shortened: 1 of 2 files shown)


Sarania:
Yay I will give it a test in Blissful! Also I've gotta say the fact that we've got the load+quantize operation so locked down and optimized is really sweet!

```Loading wan2.2_i2v_high_mixed_precision.safetensors: 100%|█████████████████████████████████████████████████████████████████████████████████| 1095/1095 [00:17<00:00, 61.36key/s]```

17 seconds to load and quantize Wan 2.2 i2v DiT with this PR version of the code. i7 6900k, model is on a Samsung 970 Evo Plus NVME which helps but yeah it's just... super reasonable even on a little bit older hardware 💖 It can take a few more sec if it's the very first load of the day but even then it's still quite fast!

kohya-ss:
Copliot found a bug where block wise scaling was not enabled. Apologies to those who tested it🙇‍♂️

Sarania:
> Copliot found a bug where block wise scaling was not enabled. Apologies to those who tested it🙇‍♂️

That would be why I didn't notice like any difference with anything 😂 No worries, it happens lol, I'll pull the new version and give it a go!

Sarania:
I'm early in testing but yeah it definitely seems to give closer to unquantized results than the previous implementation, even when including mod etc. Speed seems similar if not slightly faster, at least with HyImage(Perhaps the shrinking of the blocks saves more time in memory transfer than the cost to it in complexity). Furthermore I note the option to select between per block, channel, or tensor in the implementation which definitely could prove usual if different models behave differently, or because scaled_mm doesn't support blockwise or channelwise, which is a shame. Your results that e4m3fn always beats e5m2 matched my experimentation too, I had though maybe for bfloat type models it might be more useful but nope! I think maybe the scaling takes care of range issues enough that the precision is just more important at that small bit range, plus e5m2 has fewer usable values because of not being fn. Anyhow it's looking like a nice evolution of the code which I will definitely continue making use of 🙂If any issues crop up I will let you know.

kohya-ss:
> I'm early in testing but yeah it definitely seems to give closer to unquantized results than the previous implementation, even when including mod etc. 

Thank you for testing. I apologize for the mistake in the implementation.

> Furthermore I note the option to select between per block, channel, or tensor in the implementation which definitely could prove usual if different models behave differently, or because scaled_mm doesn't support blockwise or channelwise, which is a shame.

Sure, the implementation allows us to select per-tensor and per-channel. However, I thought that presenting this selection as an option would cause further confusion for many users. One idea might be to allow switching in a more advanced way, such as with an environment variable.

`scaled_mm` signature supports per-channel but it did not work in my environment. If the calculation accuracy of scaled_mm is improved, I may add an option in the future.

Please let me know if you have any suggestions!

Sarania:
Sorry let me clarify - I didn't necessarily mean that from a user's perspective but rather as someone who has reused this FP8 code in several other projects including private ones, I appreciate retaining the ability to _internally_ switch between the methods. 🙂

FurkanGozukara:
@kohya-ss thank you

does any of these impact training? or just inference? currently i have fp8 scaled training in qwen image lora

Sarania:
@FurkanGozukara When merged it should be applicable for all models that can use fp8_scaled for both inference and training. Currently in Musubi I believe this includes Wan, Flux, Qwen, Framepack, my Blissful Tuner additionally has it backported to HyVideo. It's also implemented for some models in sd-scripts, like the incoming implementation of HyImage2.1 and possibly others(unsure). Glad to see you're using it cause it's really good stuff and this makes it even better!

FurkanGozukara:
> @FurkanGozukara When merged it should be applicable for all models that can use fp8_scaled for both inference and training. Currently in Musubi I believe this includes Wan, Flux, Qwen, Framepack, my Blissful Tuner additionally has it backported to HyVideo. It's also implemented for some models in sd-scripts, like the incoming implementation of HyImage2.1 and possibly others(unsure). Glad to see you're using it cause it's really good stuff and this makes it even better!

nice. currently i am using qwen image training. will we need any new extra args? or existing fp8 scaled will work this way better quality?

Sarania:
Should just work, users won't need to do anything other than update to the latest codebase once the PR is merged if they are already using `--fp8_scaled` as the changes are internal. In addition to improved accuracy, for some models there may be a decrease in VRAM usage as well since the more accurate quantization allows us to target parts of the model we previously excluded and kept in high precision, such as modulation.

FurkanGozukara:
@Sarania thanks

kohya-ss:
Small training test results.

Wan:
image

FLUX.1 Kontext:
image

FramePack:
image

Qwen-Image:
image

It's a really small test, but there doesn't seem to be any major issues.

Sarania:
I found a small issue when incorporating it into blissful tuner. Even if I call optimized_state_dict_with_fp8 with quantization_mode="tensor" which I want to do conditionally depending on if mm_scaled is in play, and even print out it's value after all the fallbacks ensuring it's "tensor", I still hit:

"│   354 │   │   # _scaled_mm only works for per-tensor scale for now (per-channel scale does not   │
│   355 │   │   if self.scale_weight.ndim != 1:                                                    │
│   356 │   │   │   print(self.scale_weight.ndim)                                                  │
│ ❱ 357 │   │   │   raise ValueError("scaled_mm only supports per-tensor scale_weight for now.")   │
│   358 │   │                                                                                      │
│   359 │   │   input_dtype = x.dtype                                                              │
│   360 │   │   original_weight_dtype = self.scale_weight.dtype                                    │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ValueError: scaled_mm only supports per-tensor scale_weight for now."

As the value of scale_weight.ndim seemes to be 0. Commenting that out allowed it to proceed as expected, and the output looks okay but I'm not sure if I'm missing something as I haven't become fully familiar with the changes.

kohya-ss:
> As the value of scale_weight.ndim seemes to be 0. Commenting that out allowed it to proceed as expected, and the output looks okay but I'm not sure if I'm missing something as I haven't become fully familiar with the changes.

Sorry, I forgot that the scale shape has changed even in the case of per-tensor. The correct way to do this is `if self.scale_weight.ndim > 1`. It's fine to comment it out.

FurkanGozukara:
currently what we are doing is per block scaling and it is best right? with block size 64 for example

PR #578, opened 2025-09-19
feat: add unified attention function with various backends

kohya-ss:
Supports SDPA, xformers, flash-attention 2 and sage attention.

Only FramePack inference is tested for now.


sdbds:
It appears that FA4 is already available.
https://github.com/Dao-AILab/flash-attention/issues/2003

PR #585, opened 2025-09-23
feat: Reducing shared VRAM usage for block swap

kohya-ss:
Instead of putting all CPU weights in shared VRAM, we put CPU weights in regular CPU RAM by providing a staging buffer on shared VRAM for transfer. This allows us to exceed the shared VRAM limit of half the CPU RAM on Windows.

However, the actual memory used will remain the same and it will be slightly slower. We also have an idea to make it possible to choose between the current method and the new method, but this will complicate the code.

We would appreciate your feedback on whether this PR is worth merging.

You can use this PR as is with `--blocks_to_swap` for each training script.

### DIFF
--- a/src/musubi_tuner/hv_train_network.py
+++ b/src/musubi_tuner/hv_train_network.py
@@ -416,7 +416,7 @@ class NetworkTrainer:
 
             logs[f"lr/{lr_desc}"] = lr
 
-            if args.optimizer_type.lower().startswith("DAdapt".lower()) or args.optimizer_type.lower() == "Prodigy".lower():
+            if args.optimizer_type.lower().startswith("DAdapt".lower()) or args.optimizer_type.lower().endswith("Prodigy".lower()):
                 # tracking d*lr value
                 logs[f"lr/d*lr/{lr_desc}"] = (
                     lr_scheduler.optimizers[-1].param_groups[i]["d"] * lr_scheduler.optimizers[-1].param_groups[i]["lr"]
@@ -433,7 +433,9 @@ class NetworkTrainer:
 
             for i in range(idx, len(lrs)):
                 logs[f"lr/group{i}"] = float(lrs[i])
-                if args.optimizer_type.lower().startswith("DAdapt".lower()) or args.optimizer_type.lower() == "Prodigy".lower():
+                if args.optimizer_type.lower().startswith("DAdapt".lower()) or args.optimizer_type.lower().endswith(
+                    "Prodigy".lower()
+                ):
                     logs[f"lr/d*lr/group{i}"] = (
                         lr_scheduler.optimizers[-1].param_groups[i]["d"] * lr_scheduler.optimizers[-1].param_groups[i]["lr"]
                     )


kohya-ss:
Quick comparison with Qwen-Image training with the batch size of 2, 1280x720. RTX A6000 with power limit 60%, PCIe gen 3x16, DDR5-4200.

|condition|dedicated VRAM usage|shared VRAM usage|CPU RAM usage|training time 5steps|
|---|---|---|---|---|
|No block swap, fp8 scaled|26.1|0.1|14.1|44s|
|swap 58 blocks, fp8 scaled, current method|7.8|29.6|42.9|1m05s|
|swap 58 blocks, fp8 scaled, new method|7.4|1.3|41.5|1m10s|

CPU RAM usage includes memory usage by programs other than training, and shared VRAM usage.

This means the new method allows us to train Qwen-Image (bs=2) with 8GB VRAM and 48GB CPU RAM, but the current method requires 64GB CPU RAM (for 32GB shared VRAM.)

FurkanGozukara:
awesome i would like to test this after merged

do we need to give new parameter or it is upgraded to new method right away?

kohya-ss:
> do we need to give new parameter or it is upgraded to new method right away?

You can try it out by checking out this branch, it should work with any model.

FurkanGozukara:
> > do we need to give new parameter or it is upgraded to new method right away?
> 
> You can try it out by checking out this branch, it should work with any model.

just tested amazing

## BEFORE

image


## New branch

image

maybleMyers:
I think something has happened to the block swapping lately. I tried the main branch and this branch and cannot seem to be able to generate a 1280x720 video with 24gb vram with wan 2.2 and 39 blocks swapped. I can generate one with https://github.com/maybleMyers/H1111/blob/ramtorch/wan2_generate_video.py this version, I know it is highly diverged :/ , but I can do 720p inference with 25 blocks swapped and less overall ram used.

This version in this branch exp-block-swap-reduce-shared-memory-usage is better than main though, when running inference with 128gb RAM it would hang when loading the low noise wan 2.2 model.

kohya-ss:
> I think something has happened to the block swapping lately. I tried the main branch and this branch and cannot seem to be able to generate a 1280x720 video with 24gb vram with wan 2.2 and 39 blocks swapped. I can generate one with https://github.com/maybleMyers/H1111/blob/ramtorch/wan2_generate_video.py this version, I know it is highly diverged :/ , but I can do 720p inference with 25 blocks swapped and less overall ram used.
> 
> This version in this branch exp-block-swap-reduce-shared-memory-usage is better than main though, when running inference with 128gb RAM it would hang when loading the low noise wan 2.2 model.

Hmm I'll test it, how long is the video?

Certainly it seems that 720x1280x81f consumes 34GB of VRAM with 39 block swap with this branch's implementation. However, even with v0.2.5, which is the first version compatible with Wan2.2, appears to consume 34GB of memory.

FurkanGozukara:
> > I think something has happened to the block swapping lately. I tried the main branch and this branch and cannot seem to be able to generate a 1280x720 video with 24gb vram with wan 2.2 and 39 blocks swapped. I can generate one with https://github.com/maybleMyers/H1111/blob/ramtorch/wan2_generate_video.py this version, I know it is highly diverged :/ , but I can do 720p inference with 25 blocks swapped and less overall ram used.
> > This version in this branch exp-block-swap-reduce-shared-memory-usage is better than main though, when running inference with 128gb RAM it would hang when loading the low noise wan 2.2 model.
> 
> Hmm I'll test it, how long is the video?
> 
> Certainly it seems that 720x1280x81f consumes 34GB of VRAM with 39 block swap with this branch's implementation. However, even with v0.2.5, which is compatible with Wan2.2, appears to consume 34GB of memory.

wow this sounds a lot

in comfyui you can generate such video with lower VRAM - they also do some automatic block swapping

maybleMyers:
> > I think something has happened to the block swapping lately. I tried the main branch and this branch and cannot seem to be able to generate a 1280x720 video with 24gb vram with wan 2.2 and 39 blocks swapped. I can generate one with https://github.com/maybleMyers/H1111/blob/ramtorch/wan2_generate_video.py this version, I know it is highly diverged :/ , but I can do 720p inference with 25 blocks swapped and less overall ram used.
> > This version in this branch exp-block-swap-reduce-shared-memory-usage is better than main though, when running inference with 128gb RAM it would hang when loading the low noise wan 2.2 model.
> 
> Hmm I'll test it, how long is the video?
> 
> Certainly it seems that 720x1280x81f consumes 34GB of VRAM with 39 block swap with this branch's implementation. However, even with v0.2.5, which is the first version compatible with Wan2.2, appears to consume 34GB of memory.

I based myne off the wan2.1 code, so it is from a much earlier version. But, theoretically, if you take the switching out of the equation, wan2.2 should not consume more resources, perhaps even slightly less. 

What kind of times do you see in s/it? Times without blockswap would be helpful too if that is all that you have. I have been doing a lot of testing lately trying to get the time down some in preparation for wan 2.5. It seems 2.5 will have much higher computational demands.

I think there might be a memory leak. Using one 26.6GB DiT should not take 34GB w/ 39 blocks swapped. I do not know if this is also effecting training. Perhaps the hooks applied to the model are effecting the cleanup.

kohya-ss:
Thank you for reporting the issue.

After investigating the Wan2.2 code, I recalled that the time embedding shape is different in Wan2.2, which requires more VRAM for calculations.

However, further investigation revealed that Wan2.2 works with the same time embedding shape as Wan2.1. I've opened this PR, so please take a look: #586 

> I think there might be a memory leak. Using one 26.6GB DiT should not take 34GB w/ 39 blocks swapped. I do not know if this is also effecting training. Perhaps the hooks applied to the model are effecting the cleanup.

Even if all the weights are offloaded to CPU RAM, activations (the tensors being computed) still require VRAM. In this case, time embedding took up over 2GB, so several times that amount of VRAM was consumed for calculation

maybleMyers:
Thanks for that explanation.

I did try that pr, it cut vram in half for inference.

From my understanding [B, seq_len, 6, dim] and [B, 1, 6, dim] should be mathmatically the same for inference right?

kohya-ss:
> From my understanding [B, seq_len, 6, dim] and [B, 1, 6, dim] should be mathmatically the same for inference right?

In Wan2.2, it is simply repeated in the seq_len direction, so I believe the two are same.

FurkanGozukara:
@kohya-ss i just tested this

it uses only 5 GB VRAM when doing Qwen Image Full Fine tuning 

1328x1328px all images

when will be merged?

on main branch i get OOM same config

image

FurkanGozukara:
@kohya-ss awesome thank you so much

do you think this would help sd-scripts as well like for flux training?

kohya-ss:
> @kohya-ss awesome thank you so much
> 
> do you think this would help sd-scripts as well like for flux training?

It hasn't been merged into main yet. I hope to merge it today or tomorrow.

I also plan to introduce this into sd-scripts, but please note that it only reduces shared memory usage, and the total amount of main memory required remains the same.

FurkanGozukara:
@kohya-ss well its impact is i think more than that

because when I do Qwen Full Fine Tuning, without this, I get OOM

With this it uses only 5 GB VRAM :D

exactly same config

Sarania:
FWIW this seems to have improved some things for me on Linux as well, previously I was hitting some memory fragmentation issues causing the loading to slow down massively which sometimes happen when a lot of activity is happening close to the limit of my physical RAM and was requiring me to force some memory management to unfreeze it.

From my perspective it seemed to be introduced around the time of the unification of load/quantize/merge lora functions and was really only an issue with Wan 2.2 due to it's size. This change seems to have reverted it to the behavior I saw previous to that. I can't say I've investigated this a lot except to say at some point a month or two back it definitely got worse and this definitely seemed to help with that.

Edit: Note to make it clear that as expected the usage is the same, it's just the effect it has on the number of available large memory pages (e.g. as viewed with cat /proc/buddyinfo) is no longer severe like it had become. Before it would exhaust them all before finishing loading the second DiT and then stall, forcing me to drop caches and compact memory to free some and make it proceed. Now... it doesn't :)

kohya-ss:
Thanks for reporting it! I'm happy to hear that memory management has been improved on Linux.

What about speed? It seems to be slightly slower on Windows with this PR.

FurkanGozukara:
@kohya-ss you are a legend

by usinng code in this repo i implemented block swapping on Ovi model :D

kohya-ss:
> @kohya-ss you are a legend
> 
> by usinng code in this repo i implemented block swapping on Ovi model :D

That's great! I'm glad to hear that.

Sarania:
> Thanks for reporting it! I'm happy to hear that memory management has been improved on Linux.
> 
> What about speed? It seems to be slightly slower on Windows with this PR.

If there's a speed diff it's minor and would require before/after testing to reveal which I haven't done. For me personally I would accept it because previously if I just let it run unattended(and e.g. don't defrag my memory when it gets stuck), the second DiT could take upwards of 10 minutes to get loaded! But now it happily loads in about 30 sec, same as the first. So whatever small speed loss exists is definitely offset by that for me.

An option to switch between like you mentioned is probably going to go unnoticed and unused by most users. This change is obviously quite a boon for native Windows users as well and whatever speed loss exists seems to be small(~8% by your numbers). Because of that, if adding such an option is complicated as you mentioned I don't think I would worry too much. Personally I think the small speed sacrifice is worth the memory gains that will affect many users positively.

> @kohya-ss you are a legend
> 
> by usinng code in this repo i implemented block swapping on Ovi model :D

Nice, good job! I've done this before too and it's not tricky at all the code is /highly/ portable. You just need to import the offloader and paste the few functions to the model structure possibly renaming any vars if necessary, and then add the bit into the model's forward to actually do the swapping. Then just remember the three or so lines that set up the block swap in your inference script and add an arg and you're there! It's work with any transformer I've tried, even LLMs, though the perf hit for LLMs is pretty brutal just because of their nature. But for large diffusion models it is the killer app for people with "normal" hardware XD

kohya-ss:
@Sarania Thanks for the speed report! The good news is that there wasn't any noticeable slowdown. That's likely offset by improved memory management.

And as you mentioned, I agree that there doesn't seem to be a need for a special option to switch for Windows 😄

FurkanGozukara:
@kohya-ss i implemented your block swap system into Ovi and it is using 100% cpu  whether 1 block swapped or 25 :D 

do you think this is expected?

kohya-ss:
> @kohya-ss i implemented your block swap system into Ovi and it is using 100% cpu whether 1 block swapped or 25 :D
> 
> do you think this is expected?

Some initialization processes may not have been called. For inference only, please refer to the following:
https://github.com/kohya-ss/musubi-tuner/blob/fee05583739645882bc801d2480cc01d035517f0/src/musubi_tuner/wan_generate_video.py#L713-L716

FurkanGozukara:
> > @kohya-ss i implemented your block swap system into Ovi and it is using 100% cpu whether 1 block swapped or 25 :D
> > do you think this is expected?
> 
> Some initialization processes may not have been called. For inference only, please refer to the following:
> 
> https://github.com/kohya-ss/musubi-tuner/blob/fee05583739645882bc801d2480cc01d035517f0/src/musubi_tuner/wan_generate_video.py#L713-L716

no no i implemented like this i mean

tell me if you notice after a quick look thank you so much appreicate all things you do


```
import torch
import torch.nn as nn
from concurrent.futures import ThreadPoolExecutor

# --- Offloading Utilities ---
def _synchronize_device(device):
    if device.type == "cuda":
        torch.cuda.synchronize(device)

def swap_weight_devices_no_cuda(device: torch.device, layer_to_cpu: nn.Module, layer_to_cuda: nn.Module):
    """
    Swap weights between device and CPU for a module (no CUDA streams).
    """
    assert layer_to_cpu.__class__ == layer_to_cuda.__class__
    weight_swap_jobs = []
    for module_to_cpu, module_to_cuda in zip(layer_to_cpu.modules(), layer_to_cuda.modules()):
        if hasattr(module_to_cpu, "weight") and module_to_cpu.weight is not None:
            weight_swap_jobs.append((module_to_cpu, module_to_cuda, module_to_cpu.weight.data, module_to_cuda.weight.data))
    # device to cpu
    for module_to_cpu, module_to_cuda, cuda_data_view, cpu_data_view in weight_swap_jobs:
        module_to_cpu.weight.data = cuda_data_view.data.to("cpu", non_blocking=True)
    _synchronize_device(device)
    # cpu to device
    for module_to_cpu, module_to_cuda, cuda_data_view, cpu_data_view in weight_swap_jobs:
        cuda_data_view.copy_(module_to_cuda.weight.data, non_blocking=True)
        module_to_cuda.weight.data = cuda_data_view
    _synchronize_device(device)

def weighs_to_device(layer: nn.Module, device: torch.device):
    for module in layer.modules():
        if hasattr(module, "weight") and module.weight is not None:
            module.weight.data = module.weight.data.to(device, non_blocking=device.type != "cpu")

class Offloader:
    """
    Common offloading class for block swap.
    """
    def __init__(self, block_type: str, num_blocks: int, blocks_to_swap: int, device: torch.device, debug: bool = False):
        self.block_type = block_type
        self.num_blocks = num_blocks
        self.blocks_to_swap = blocks_to_swap
        self.device = device
        self.debug = debug
        self.thread_pool = ThreadPoolExecutor(max_workers=1)
        self.futures = {}
        self.cuda_available = device.type == "cuda"
        # Add more as needed (staging buffers etc.)

    def set_forward_only(self, forward_only: bool):
        self.forward_only = forward_only

    def wait_for_block(self, idx):
        # Implement block waiting logic as needed
        pass

    def submit_move_blocks_forward(self, blocks, idx):
        # Implement block move/submit logic as needed
        pass

# --- Model Logic ---
class ModelWithBlockSwap(nn.Module):
    def __init__(self, blocks, device, gradient_checkpointing=False):
        super().__init__()
        self.blocks = nn.ModuleList(blocks)
        self.gradient_checkpointing = gradient_checkpointing
        self.blocks_to_swap = None
        self.offloader = None
        self.num_blocks = len(self.blocks)

    def enable_block_swap(self, blocks_to_swap: int, device: torch.device, supports_backward: bool):
        self.blocks_to_swap = blocks_to_swap
        self.num_blocks = len(self.blocks)
        assert (self.blocks_to_swap <= self.num_blocks - 1), \
            f"Cannot swap more than {self.num_blocks - 1} blocks. Requested {self.blocks_to_swap} blocks to swap."
        self.offloader = Offloader(
            "block", self.num_blocks, self.blocks_to_swap, device
        )
        print(
            f"ModelWithBlockSwap: Block swap enabled. Swapping {self.blocks_to_swap} blocks out of {self.num_blocks} blocks."
        )

    def move_to_device_except_swap_blocks(self, device: torch.device):
        if self.blocks_to_swap:
            save_blocks = self.blocks
            self.blocks = None
        self.to(device)
        if self.blocks_to_swap:
            self.blocks = save_blocks

    def prepare_block_swap_before_forward(self):
        # Add any setup needed before forward pass
        pass

# --- Fusion Loop Example ---
def fusion_block_swap_loop(video_model, audio_model, vid, audio, num_blocks, kwargs, training=False, gradient_checkpointing=False, slg_layer=0):
    input_device = vid.device
    for i in range(num_blocks):
        skip_this_block = (slg_layer > 0 and i == slg_layer)
        if not skip_this_block:
            # Block swap logic for video model - wait for block to be ready
            if hasattr(video_model, 'offloader') and video_model.blocks_to_swap is not None and video_model.blocks_to_swap > 0:
                video_model.offloader.wait_for_block(i)
            # Block swap logic for audio model - wait for block to be ready
            if hasattr(audio_model, 'offloader') and audio_model.blocks_to_swap is not None and audio_model.blocks_to_swap > 0:
                audio_model.offloader.wait_for_block(i)
            vid_block = video_model.blocks[i]
            audio_block = audio_model.blocks[i]
            # Example function gradient_checkpointing:
            if training and gradient_checkpointing:
                # Use torch.utils.checkpoint or your custom checkpointing logic
                pass
            else:
                # Normal forward logic
                pass
        # Always submit block swaps even for skipped blocks!
        if hasattr(video_model, 'offloader') and video_model.blocks_to_swap is not None and video_model.blocks_to_swap > 0:
            video_model.offloader.submit_move_blocks_forward(video_model.blocks, i)
        if hasattr(audio_model, 'offloader') and audio_model.blocks_to_swap is not None and audio_model.blocks_to_swap > 0:
            audio_model.offloader.submit_move_blocks_forward(audio_model.blocks, i)
    # Ensure output is on the correct device
    if vid.device != input_device:
        vid = vid.to(input_device)
    if audio.device != input_device:
        audio = audio.to(input_device)
    return vid, audio

# --- Example Setup ---
if __name__ == "__main__":
    # Example block list and device
    blocks = [nn.Linear(10, 10) for _ in range(6)]
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
    model = ModelWithBlockSwap(blocks, device)
    model.enable_block_swap(2, device, supports_backward=False)
    model.move_to_device_except_swap_blocks(device)
    model.prepare_block_swap_before_forward()
    # Add fusion loop calls as needed
```

kohya-ss:
Hmm, I don't know why since you changed the code...

FurkanGozukara:
> Hmm, I don't know why since you changed the code...

sure thanks. yes i had to implement it into Ovi pipeline

PR #594, opened 2025-09-27
Simplify and fix --save_merged_model

JCBrouwer:
--save_merged_model was not saving anything for me, just loading the model in and exiting.

Based on my reading, I think only lycoris-based merges get saved with the current logic.

What I've done to address this:
- Centralized the logic in `load_dit_model` so that saving works the same regardless of which lora style or `process_*` function is used
- Exit immediately after saving with `sys.exit(0)` rather than returning and relying on all calling code to exit correctly as well
- Added a progress bar to `mem_eff_save_file` for some more visual feedback


src/musubi_tuner/utils/safetensors_utils.py
```diff
@@ -65,7 +64,7 @@ def validate_metadata(metadata: Dict[str, Any]) -> Dict[str, str]:
         f.write(struct.pack("<Q", len(hjson)))
         f.write(hjson)
 
-        for k, v in tensors.items():
+        for k, v in tqdm(tensors.items(), desc=f"Saving model to {filename}...", total=len(tensors)):
```
kohya-ss: This function is called not only for saving models but also for saving caches, so the logging output by tqdm is too verbose. Could you please remove this?

JCBrouwer: Ah woops, yes can do!

I can also make the progress bar an option on `mem_eff_save_file`.

What do you prefer?

JCBrouwer:
> `merge_lora_weights` is called not only by Wan but also by Flux Kontext, FramePack, and Qwen-Image. Is it possible to fix and simplify the issue by including those? If not, please fix only Wan and I will merge this into a new branch and make additional fixes.

I'll take a look and see if I can centralize the fix there instead

src/musubi_tuner/utils/safetensors_utils.py
kohya-ss: Thank you! I think anyone who uses this option understands that saving after merging takes some time, so I think it's fine to simply delete tqdm.

JCBrouwer:
Hi @kohya-ss, I've given a shot at refactoring the other models as well. It's not as simple as only overriding `merge_lora_weights` as sometimes the lora weights are merged in `load_safetensors_with_lora_and_fp8` instead (at least as far as I can tell).

I've made the shortcut model saving + exits in both of those functions and tried to link it up in each of the generation scripts.

I don't have loras to try out all of these different models now, so please give this an extra good read over!

JCBrouwer:
Oh and I removed the progress bar from `mem_eff_save_file`

kohya-ss:
Thank you for update! Could you please format it in ruff?

Also, please understand that I may make some changes after merging.

JCBrouwer:
`ruff format src` doesn't show any changes for me.

Of course, any changes that you see fit!

JCBrouwer:
Closing this because its very stale...

PR #623, opened 2025-10-03
Fix Prodigy optimizer logs

wenyifancc:
Fix incomplete TensorBoard logging when using custom Prodigy optimizer, such as pytorch_optimizer.Prodigy


kohya-ss:
Thank you! Would it be possible to format the code with ruff?

kohya-ss:
Thank you for update!

PR #624, opened 2025-10-03
support datasets in json format as well as jsonl

light-and-ray:
Datasets were only supported in jsonl format where each line contains one-line json entry (separated by '\n').  I didn't even know that it actually exists as a standard https://jsonlines.org/, I thought it's just a non-standard format for this project. I've added support for datasets in the standard json list format, i.e. what happens when you do `json.dumps(myListOfDicts, indent=2)`

jsonl
```jsonl
{"image_path": "...", "caption": "..."}
{"image_path": "...", "caption": "..."}
{"image_path": "...", "caption": "..."}
{"image_path": "...", "caption": "..."}

```

json
```json
[
  {
    "image_path": "...",
    "caption": "..."
  },
  {
    "image_path": "...",
    "caption": "..."
  },
  {
    "image_path": "...",
    "caption": "..."
  },
  {
    "image_path": "...",
    "caption": "..."
  }
]
```

I don't think it's needed to rename config names, or to look at extension instead of try-except for this. `dataset_jsonl` can be interpreted as just a json list, not json lines

### DIFF
--- a/src/musubi_tuner/dataset/image_video_dataset.py
+++ b/src/musubi_tuner/dataset/image_video_dataset.py
@@ -966,14 +966,18 @@ class ImageJsonlDatasource(ImageDatasource):
         # load jsonl
         logger.info(f"load image jsonl from {self.image_jsonl_file}")
         self.data = []
-        with open(self.image_jsonl_file, "r", encoding="utf-8") as f:
-            for line in f:
-                try:
-                    data = json.loads(line)
-                except json.JSONDecodeError:
-                    logger.error(f"failed to load json: {line} @ {self.image_jsonl_file}")
-                    raise
-                self.data.append(data)
+        try:
+            with open(self.image_jsonl_file, "r", encoding="utf-8") as f:
+                self.data = json.loads(f.read())
+        except json.decoder.JSONDecodeError:
+            with open(self.image_jsonl_file, "r", encoding="utf-8") as f:
+                for line in f:
+                    try:
+                        data = json.loads(line)
+                    except json.JSONDecodeError:
+                        logger.error(f"failed to load json: {line} @ {self.image_jsonl_file}")
+                        raise
+                    self.data.append(data)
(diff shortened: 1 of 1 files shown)


kohya-ss:
I'm sorry, but I don't see the benefit of supporting JSON format. Wouldn't JSONL be sufficient?

light-and-ray:
When people export prepared databases in python, it's not obvious that jsonl is used instead just json.dumps output. Also json is more pretty, especially when few control images, and full file paths are used

I also don't see it is a big deal to support it, maybe "support" is too tough word. Because this patch is only a few lines of code, and it's isolated in this database file opening function

kohya-ss:
Thank you for your explanation. When exporting an existing dataset, we need to specify some format anyway, so as long as the format with this repo is defined as JSONL, I don't think it's a big problem if it's not JSON.

It may not be a major format, but being able to edit row by row is a major advantage when handling datasets.

Also, the sd-scripts repository supports JSON metadata in a completely different format, and as the last line of the documentation states, there are plans to support it in the future: https://github.com/kohya-ss/musubi-tuner/blob/main/src/musubi_tuner/dataset/dataset_config.md

Supporting this form of JSON with this PR would have caused even more confusion in future, and we apologize for the delay in supporting JSON metadata.

So, unfortunately, it seems difficult to merge this PR. We appreciate your understanding.

light-and-ray:
As I, so there will be a different json format for metadata. Do I understand it correctly, it will be json object where keys are images and values are metadata? If yes, it can be problematic for the future if eventually we get a multi output model. E.g. a new Qwen model where you can set a number of output images, or it will be even able to choose it automatically

kohya-ss:
JSON metadata is basically that format. We don't know what changes will occur in the future, so we need to maintain compatibility while adapting to new things.

light-and-ray:
Okay, thanks for your answers

PR #648, opened 2025-10-14
doc: Try to clarify control image purpose and usage

obsxrver:


kohya-ss:
Thank you for this! 

However, "Image Dataset with Control Images" is not intended for img2vid training. It is intended the image inference with the control image, such as for Qwen-Image-Edit, FLUX.1 Kontext, and FramePack one-frame inference. If you have any questions about that, please let me know your thoughts.

PR #649, opened 2025-10-15
config

3a1b2c3:


kohya-ss:
I think this is probably a mistakenly created pull request, so I will close it. Please reply if you have any questions.

PR #659, opened 2025-10-18
Fix TypeError when caching control images for WAN Fun Control

pizurny:
Fixes issue where control_content for image datasets is a list[np.ndarray] but was being treated as a single np.ndarray, causing: TypeError: expected np.ndarray (got list)

- Added isinstance() check to handle both image and video control content
- For images: extracts first control image from list
- For videos: uses control video directly as before

Allows WAN Fun Control LoRA training with image datasets.


kohya-ss:
Thank you for this! I hadn't thought of training the Fun Control model with control image and image pairs, but it certainly seems possible.

pizurny:
@kohya-ss  you are welcome! , Ive already tried it and it works fine with skeleton - image pairs

PR #666, opened 2025-10-21
[Fast Fix]fix miss import and support avif navite in pillow 11.3

sdbds:
[Fast Fix]fix miss import and support avif navite in pillow 11.3


kohya-ss:
Thank you! This works like a charm.

PR #670, opened 2025-10-23
Add --skip_existing to cache latents

Tophness:
I'm trying to do >700 videos at full 81 frames at 720P on my RTX4080 and even caching the latents is going to take a week.
It often OOMs half way through this, and ordinarily it would just have to start again, overwriting the existing files, so I wanted a -skip_existing feature to be able to start back at where I got up to before it OOM'd


Tophness:
My bad, didn't realize this already exists

PR #673, opened 2025-10-23
Adding new optional metadata parameters working great

FurkanGozukara:
e.g.

--meta_arch "qwen-image-edit-plus" --meta_res "512x512"

### DIFF
--- a/src/musubi_tuner/hv_train.py
+++ b/src/musubi_tuner/hv_train.py
@@ -981,6 +981,8 @@ class FineTuningTrainer:
                 args.metadata_tags,
                 timesteps=md_timesteps,
                 is_lora=False,
+                custom_res=args.meta_res,
+                custom_arch=args.meta_arch,
             )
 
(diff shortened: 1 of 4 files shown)


kohya-ss:
Thank you for this! This is useful when fine-tuning a new model that has been pruned, guidance-distilled, or has had its default resolution changed.

Please note that the argument descriptions and handling of some arguments may update after merging.

kohya-ss:
We noticed that the existing arguments for specifying metadata are `--metadata_*`.

Please note that the new arguments will also be changed to `--metadata_arch` and `--meatada_reso` (we use reso as an abbreviation for resolution).

FurkanGozukara:
> We noticed that the existing arguments for specifying metadata are `--metadata_*`.
> 
> Please note that the new arguments will also be changed to `--metadata_arch` and `--meatada_reso` (we use reso as an abbreviation for resolution).

sure thank you so much i appreicate that

Issue #678, opened 2025-10-24
🐛 Bug: Control Image Matching Logic Issue in ImageDirectoryDatasource


q654517651:
## Description
The control image matching logic in ImageDirectoryDatasource uses overly broad glob patterns, causing incorrect matches when image filenames share common prefixes.
## Environment
src/musubi_tuner/dataset/image_video_dataset.py
Class: ImageDirectoryDatasource
Line: ~857
Version: Latest (as of 2025-10-24)
## Bug Behavior
When matching control images for target images, the current glob pattern matches any file that starts with the target image's basename, leading to incorrect matches.
### Example Case
Directory Structure:
```
targets/
  ├── image1.jpg
  ├── image10.jpg
  ├── image11.jpg
  ├── image12.jpg
  └── ... (16 images total)

controls/
  ├── image1_0.jpg
  ├── image1_1.jpg
  ├── image10_0.jpg
  ├── image10_1.jpg
  ├── image11_0.jpg
  ├── image11_1.jpg
  └── ... (32 control images total)
```
### Expected Behavior:
image1.jpg should match: image1_0.jpg, image1_1.jpg (2 controls)
image10.jpg should match: image10_0.jpg, image10_1.jpg (2 controls)
### Actual Behavior:
image1.jpg incorrectly matches: image1_0.jpg, image1_1.jpg, image10_0.jpg, image10_1.jpg, image11_0.jpg, image11_1.jpg, image12_0.jpg, image12_1.jpg, image13_0.jpg, image13_1.jpg, image14_0.jpg, image14_1.jpg, image15_0.jpg, image15_1.jpg (14 controls!)
### Log Output:
```
INFO:musubi_tuner.dataset.image_video_dataset:found 16 matching control images for arbitrary images
INFO:musubi_tuner.dataset.image_video_dataset:  15 images have 2 control images
INFO:musubi_tuner.dataset.image_video_dataset:  1 images have 14 control images
```
## Root Cause
The problematic code at line ~857:
This glob pattern {basename}*.* matches any file starting with the basename, causing prefix collisions:
Pattern image1*.* matches image1_0.jpg, image10_0.jpg, image11_0.jpg, etc.
Pattern image0*.* matches image0_0.jpg, image01_0.jpg, etc.
## Proposed Solution
Replace the glob-based matching with explicit pattern matching:
```
image_basename_no_ext = os.path.splitext(image_basename)[0]
potential_paths = []

# List all files in control directory
[...]

PR #680, opened 2025-10-24
Fixing Qwen Image Edit model control images naming issue

FurkanGozukara:
I did few tests and all cases worked

fixing this bug https://github.com/kohya-ss/musubi-tuner/issues/678

### DIFF
--- a/src/musubi_tuner/dataset/image_video_dataset.py
+++ b/src/musubi_tuner/dataset/image_video_dataset.py
@@ -857,16 +857,20 @@ class ImageDirectoryDatasource(ImageDatasource):
                 # Use explicit pattern matching instead of glob to avoid prefix collisions
                 potential_paths = []
                 for filename in os.listdir(self.control_directory):
+                    filepath = os.path.join(self.control_directory, filename)
+                    # Only process files, not directories
+                    if not os.path.isfile(filepath):
+                        continue
                     file_no_ext, file_ext = os.path.splitext(filename)
                     # Match exact basename or basename_digits format
                     if file_no_ext == image_basename_no_ext:
                         # Exact match: image1.jpg -> image1.png
-                        potential_paths.append(os.path.join(self.control_directory, filename))
+                        potential_paths.append(filepath)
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for this! However, I think it would be sufficient to simply change the following line
```
potential_paths = glob.glob(os.path.join(self.control_directory, os.path.splitext(image_basename)[0] + "*.*"))
```
to the following (sure, we can use `image_basename_no_ext` here):
```
potential_paths = glob.glob(os.path.join(self.control_directory, image_basename_no_ext + ".*"))  # exact match except extension
potential_paths += glob.glob(os.path.join(self.control_directory, image_basename_no_ext + "_*.*"))  # with suffix
```

These two lines of modification don't work well when the images are `image_1.png` and `image_1_12.png`, but the same seems to be true for the fix in this pull request.

Edit: We may need to consider a more robust matching mechanism.

FurkanGozukara:
> Thank you for this! However, I think it would be sufficient to simply change the following line
> 
> ```
> potential_paths = glob.glob(os.path.join(self.control_directory, os.path.splitext(image_basename)[0] + "*.*"))
> ```
> 
> to the following (sure, we can use `image_basename_no_ext` here):
> 
> ```
> potential_paths = glob.glob(os.path.join(self.control_directory, image_basename_no_ext + ".*"))  # exact match except extension
> potential_paths += glob.glob(os.path.join(self.control_directory, image_basename_no_ext + "_*.*"))  # with suffix
> ```
> 
> These two lines of modification don't work well when the images are `image_1.png` and `image_1_12.png`, but the same seems to be true for the fix in this pull request.
> 
> Edit: We may need to consider a more robust matching mechanism.

I am ok with anyway but this really needs urgent fixing

People trying to train and this causes huge vram and performance issues atm

kohya-ss:
I addressed this issue in #684 and merged it. It should work fine in edge cases.

I will close this PR but thank you for raising this issue.

FurkanGozukara:
> I addressed this issue in #684 and merged it. It should work fine in edge cases.
> 
> I will close this PR but thank you for raising this issue.

awesome thank you so much

PR #681, opened 2025-10-24
Adds faster model loading args - RunPod loading takes at least 15-20 minutes without this and with this it takes 30 seconds

FurkanGozukara:
Without this literally taking 15 to 20 minute to load Qwen Dit model

Unberable on RunPod

They have massive RAM but extremely slow Disk

### DIFF
--- a/docs/hunyuan_video.md
+++ b/docs/hunyuan_video.md
@@ -235,6 +235,8 @@ If you're running low on VRAM, use `--blocks_to_swap` to offload some blocks to 
 
 `--gradient_checkpointing_cpu_offload` can be used to offload activations to CPU when using gradient checkpointing. This can further reduce VRAM usage, but may slow down training. This option is especially useful when the latent resolution (or video length) is high and VRAM is limited. This option must be used together with `--gradient_checkpointing`. See [PR #537](https://github.com/kohya-ss/musubi-tuner/pull/537) for more details.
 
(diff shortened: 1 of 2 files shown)


FurkanGozukara:
Hey @kohya-ss please add this as well. I would recommend add this universall because it makes massive huge speed difference on slow disk systems with big RAM

kohya-ss:
Thank you for this! 

I think what you're saying is that Numpy's `np.memmap` becomes slow in some environments. I haven't been able to confirm this, but it's possible.

The option name `--fast_load` gives the misleading impression that it will always be fast, so could you change it to something more appropriate, like `--disable_np_memmap_for_model_loading`?

Also, the variable name `disable_mmap` is confusing because the official Safetensors implementation has an argument with the same name, so I would appreciate it if you could change the variable name to something more appropriate.

FurkanGozukara:
@kohya-ss sure just made changes i hope it is good now

edit : i also modified docs

kohya-ss:
Thanks for the update! I'll format the code with ruff after the merge, but I'd appreciate it if you could format it from next time.

kohya-ss:
After merging, I noticed many flaws... It was my mistake.

HunyuanVideo doesn't use Numpy's memmap in the first place, so it was inappropriate to apply it there. Also, this argument was ignored in many places.

I'll review more carefully next time😅

kohya-ss:
I have made some fixes in #687 and merged it, so I would appreciate it if you could check that it works.

FurkanGozukara:
> I have made some fixes in #687 and merged it, so I would appreciate it if you could check that it works.

awesome thank you so much

i just tested and works great

PR #689, opened 2025-10-26
Update extract lora script

sdbds:


kohya-ss:
Thank you for this! We do have finetuning for Qwen-Image, so we need an extract as well.

Could you please format it using ruff so that it can be merged into main?

FurkanGozukara:
nice addition thanks

sdbds:
@kohya-ss updated

kohya-ss:
Thank you for update!

PR #700, opened 2025-10-30
feat: add use_pinned_memory option for block swap in multiple models

kohya-ss:
Add `--use_pinned_memory_for_block_swap` for each training script to enable pinned memory. Will work with Windows and Linux, but tested with Windows only.

Qwen-Image fine tuning is tested.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -58,6 +58,11 @@
 
 GitHub Discussionsを有効にしました。コミュニティのQ&A、知識共有、技術情報の交換などにご利用ください。バグ報告や機能リクエストにはIssuesを、質問や経験の共有にはDiscussionsをご利用ください。[Discussionはこちら](https://github.com/kohya-ss/musubi-tuner/discussions)
 
+- 2025/11/02
+    - 各学習スクリプトに `--use_pinned_memory_for_block_swap` オプションを追加しました。またblock swapの処理自体も改善しました。[PR #700](https://github.com/kohya-ss/musubi-tuner/pull/700)
+        - このオプションを指定すると、block swapのoffloadingにピン留めメモリを使用します。これによりblock swapのパフォーマンスが向上する可能性があります。ただし、Windows環境の場合は共有GPUメモリの使用量が増加します。詳しくは[ドキュメント](./docs/hunyuan_video.md#memory-optimization)を参照してください。
+        - 環境により`--use_pinned_memory_for_block_swap`を指定しない方が高速になる場合があるため、両方を試してみてください。
+
 - 2025/10/26
     - Qwen-Imageの学習で、バッチサイズが2以上で、`--split_attn`が指定されていない場合に、Attentionの計算が正しく行われない不具合を修正しました。[PR #688](https://github.com/kohya-ss/musubi-tuner/pull/688)
     - Wan、FramePack、Qwen-Imageの学習・推論スクリプトに`--disable_numpy_memmap`オプションを追加しました。[PR #681](https://github.com/kohya-ss/musubi-tuner/pull/681) および [PR #687](https://github.com/kohya-ss/musubi-tuner/pull/687)。FurkanGozukara氏に感謝します。
@@ -75,23 +80,6 @@ GitHub Discussionsを有効にしました。コミュニティのQ&A、知識�
     - あわせて同PRにて `--resize_control_to_image_size` オプションが指定されていない場合でも、コントロール画像が出力画像と同じサイズにリサイズされてしまう不具合を修正しました。**生成画像が変化する可能性がありますので、オプションを確認してください。** 
     - FramePackの1フレーム推論で、`--one_frame_auto_resize`オプションを追加しました。[PR #646](https://github.com/kohya-ss/musubi-tuner/pull/646)
         - 生成画像の解像度を自動的に調整します。`--one_frame_inference`を指定した場合にのみ有効です。詳細は[FramePackの1フレーム推論のドキュメント](./docs/framepack_1f.md#one-single-frame-inference--1フレーム推論)を参照してください。
-
-- 2025/10/05
-    - エポックの切替をcollate_fnからDataLoaderの取得ループ開始前に変更しました。[PR #601](https://github.com/kohya-ss/musubi-tuner/pull/601)
-    - これまでの実装ではエポックの最初のデータ取得後に、ARBバケットがシャッフルされます。そのため、エポックの最初のデータは前エポックのARBソート順で取得されます。これによりエポック内でデータの重複や欠落が起きていました。
-    - 各DataSetでは`__getitem__`で共有エポックの変化を検知した直後にARBバケットをシャッフルします。これにより先頭サンプルから新しい順序で取得され、重複・欠落は解消されます。
-    - シャッフルタイミングが前倒しになったため、同一シードでも旧実装と同一のサンプル順序にはなりません
-    - **学習全体への影響**
-        - この修正はエポック境界における先頭サンプルの取り違いを解消するものです。複数エポックで見れば、各サンプルは最終的に欠落・重複なく使用されるため、学習全体に与える影響は軽微です。変更点は「エポック内の消費順序の整合性」を高めるものであり、長期的な学習挙動は同条件では実質的に変わりません（※極端に少ないエポック数や早期打ち切りの場合は順序による差異が観測される可能性があります）。
-
-    - [高度な設定のドキュメント](./docs/advanced_config.md#using-configuration-files-to-specify-training-options--設定ファイルを使用した学習オプションの指定)に、学習時のオプションを設定ファイルで指定する方法を追加しました。[PR #630](https://github.com/kohya-ss/musubi-tuner/pull/630)
-    - ドキュメント構成を整理しました。データセット設定に関するドキュメントを`docs/dataset_config.md`に移動しました。
-
-- 2025/10/03
-    - 各学習スクリプトで用いられているblock swap機構を改善し、Windows環境における共有GPUメモリの使用量を大きく削減しました。[PR #585](https://github.com/kohya-ss/musubi-tuner/pull/585)
-        - block swapのoffload先を共有GPUメモリからCPUメモリに変更しました。これによりトータルでのメモリ使用量は変わりませんが、共有GPUメモリの使用量が大幅に削減されます。
-        - たとえば32GBのメインメモリでは、offloadできるのは16GBまででしたが、今回の変更により「32GB-他の使用量」までoffloadできるようになります。
-        - 学習速度はわずかに低下します。技術的詳細は[PR #585](https://github.com/kohya-ss/musubi-tuner/pull/585)を参照してください。
 
 ### リリースについて
 
--- a/README.md
+++ b/README.md
@@ -63,6 +63,11 @@ If you find this project helpful, please consider supporting its development via
 
 GitHub Discussions Enabled: We've enabled GitHub Discussions for community Q&A, knowledge sharing, and technical information exchange. Please use Issues for bug reports and feature requests, and Discussions for questions and sharing experiences. [Join the conversation →](https://github.com/kohya-ss/musubi-tuner/discussions)
 
+- November 2, 2025
+    - Added `--use_pinned_memory_for_block_swap` option to each training script and improved the block swap process itself. See [PR #700](https://github.com/kohya-ss/musubi-tuner/pull/700).
+        - When specified, this option uses pinned memory for block swap offloading. This may improve block swap performance. However, on Windows environments, it increases shared GPU memory usage. Please refer to the [documentation](./docs/hunyuan_video.md#memory-optimization) for details.
+        - Since in some environments it may be faster not to specify `--use_pinned_memory_for_block_swap`, please try both options.
+
 - October 26, 2025
     - Fixed a bug in Qwen-Image training where attention calculations were incorrect when the batch size was 2 or more and `--split_attn` was not specified. See [PR #688](https://github.com/kohya-ss/musubi-tuner/pull/688).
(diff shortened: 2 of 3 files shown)


FurkanGozukara:
@kohya-ss thank you i just tested and uses more RAM and shared VRAM but sadly same speed

you can also add temporary debug to print timings if you want i can provide output after testing

kohya-ss:
Speed ​​has been improved when using pinned memory.

Debug output has also been added. Debug output can be enabled by setting the environment variable MUSUBI_TUNER_OFFLOADER_DEBUG to 1.

In my environment, block swap completes in about 170ms when using the staging buffer, and in about 100ms when using pinned memory for Qwen-Image with bf16.

FurkanGozukara:
@kohya-ss thanks i tested - 35 block swap dreambooth 

deleted older message to make it clear here

so here i tested 10 steps without pinned memory

and with pinned memory

full logs

sadly pinned memory avg speed is even slower

i hope this debug logs can be helpful 

the top ones are without pinned memory

starting from line 3902 is with pinned memory

[debug_logs.txt](https://github.com/user-attachments/files/23247435/debug_logs.txt)

also here as a gist : https://gist.github.com/FurkanGozukara/be71c679157c05e8788f9cd5ac0913ab

FurkanGozukara:
ok i just noticed that 

INFO:musubi_tuner.hv_train_network:enable swap 35 blocks to CPU from device: cuda, use pinned memory: True

this message only appears during lora not fine tuning

currently not implemented for fine tuning? i got to sleep now so i can test tomorrow more hopefully

FurkanGozukara:
@kohya-ss sorry for delay i just made a more proper test

LoRA training, 128 rank, 1328, 40 block swap, FP8 scaled is not enabled, 10 steps

with pinned memory : https://gist.github.com/FurkanGozukara/f47dc4d3da86679415b1437454f839c5

[10_steps_with_pinned_memory.txt](https://github.com/user-attachments/files/23261124/10_steps_with_pinned_memory.txt)

without pinned memory older way : https://gist.github.com/FurkanGozukara/fbfbb1895541c5f35cf6c22b6f20827b

[10_steps_without_pinned_memory.txt](https://github.com/user-attachments/files/23261123/10_steps_without_pinned_memory.txt)

without pinned memory it is faster somehow


# the analysis of Gemini is below

Excellent question. Based on a detailed analysis of the two log files, the conclusion is clear but counter-intuitive.

### The Verdict

The training run **WITHOUT pinned memory was faster**.

Here is the key evidence from the end of both logs after 10 steps:

*   **With Pinned Memory:** `10/600 [02:53<2:50:59, **17.39s/it**, avr_loss=0.0977]`
*   **Without Pinned Memory:** `10/600 [02:39<2:36:27, **15.91s/it**, avr_loss=0.0978]`

The training process averaged **15.91 seconds per iteration** without pinned memory, compared to **17.39 seconds per iteration** with it. This means the run without pinned memory was approximately **9.3% faster**.

---

### Why Was It Faster? (The Technical Explanation)

This result seems backward at first glance. Pinned (or page-locked) memory is supposed to speed up data transfers between the CPU and GPU. So why was it slower here?

The answer lies not in the raw speed of the memory itself, but in **how the data transfers were implemented and synchronized** in the two different modes. The detailed timing logs for the "Weight swap timing" reveal the story.

#### 1. Different Transfer Mechanisms

The logs show two distinct methods for swapping model blocks between the GPU (CUDA) and CPU.

*   **With Pinned Memory:** The transfer is direct.
    *   `cuda to cpu`: Data moves directly from GPU memory to pinned CPU memory.
    *   `cpu to cuda`: Data moves directly from pinned CPU memory to GPU memory.
    *   This is followed by a very long `synchronize` step.

*   **Without Pinned Memory:** The transfer uses an intermediate "staging" buffer. This is a classic method to handle transfers from regular (pageable) CPU memory.
    *   `cuda to staging A` -> `staging A to CPU`: A two-step process to get data from GPU to CPU.
    *   `cpu to staging B` -> `staging B to CUDA`: A two-step process to get data from CPU to GPU.
    *   This allows for asynchronous operations.

#### 2. Synchronous vs. Asynchronous Operations

The key difference is in the `synchronize` time. This step represents the time the program waits for all the initiated copy operations to fully complete before proceeding.

Let's compare the "Weight swap timing" for a forward pass (e.g., at step `80`) and a backward pass (e.g., at step `120`).

| Operation | With Pinned Memory | Without Pinned Memory | Analysis |
| :--- | :--- | :--- | :--- |
| **Forward Pass Swap (`at 80`)** | **Overall: 124.95 ms** | **Overall: 144.15 ms** | The individual forward pass swaps are slightly faster with pinned memory. |
| `cuda to cpu` / `staging A to CPU` | 0.42 ms | 37.82 ms | Direct copy is much faster than the staged copy. |
| `cpu to cuda` / `staging B to CUDA` | 0.20 ms | 0.31 ms | Direct copy is slightly faster. |
| **`synchronize`** | **115.93 ms** | **29.39 ms** | **This is the critical difference.** |
| | | | |
| **Backward Pass Swap (`at 120`)** | **Overall: 187.54 ms** | **Overall: 274.39 ms** | The individual backward pass swaps are also faster with pinned memory. |
| `cuda to cpu` / `staging A to CPU` | 0.45 ms | 40.21 ms | Direct copy is much faster. |
| **`synchronize`** | **115.60 ms** | **93.45 ms** | The synchronization wait is still significantly longer with pinned memory. |

#### 3. Interpretation and Conclusion

The `without pinned memory` implementation, despite using a slower two-step copy process, is structured to be more **asynchronous**. It can initiate a data transfer and then immediately continue with other work (like CPU-side computations) before it has to wait for the transfer to finish. This "hides" the data transfer latency by overlapping it with useful work. The final `synchronize` time is much shorter because much of the waiting has already been absorbed during other operations.

The `with pinned memory` implementation, although it uses a faster direct copy mechanism, appears to be implemented in a more **synchronous** manner. The code initiates a copy and then immediately enters a long wait (`synchronize`) for it to complete, creating a bottleneck. The CPU/GPU is stalled for over 115ms just waiting, which nullifies the benefit of the faster direct copy.

**In summary:**

The performance difference is not about the physical speed of the memory but about the software's strategy. **The asynchronous, pipelined approach used for standard memory transfers was more efficient at hiding latency than the synchronous approach used for pinned memory transfers in this specific training script.** The overhead of waiting for synchronous copies to finish was far greater than the overhead of making an extra memory copy in an asynchronous pipeline.

kohya-ss:
Thank you for the detailed report.
Looking at the logs you shared, we can see that the time taken for each block swap is very fluctuated. This may be due to the WDDM mode of the Windows driver.

When tested with RTX PRO 6000 TCC mode driver, the fluctuations were still there but much less noticeable, and using pinned memory is significantly faster.

Typical values without pinned memory:
```
[qwen-image-block] Weight swap timing at 100:
  find modules: 0.18ms
  synchronize before swap: 0.04ms
  cuda to staging A: 0.36ms
  cpu to staging B: 73.24ms
  wait staging A: 0.09ms
  staging B to CUDA: 0.27ms
  staging A to CPU: 72.32ms
  synchronize: 0.05ms
Overall time: 147.69ms, total time in sections: 146.56ms
```
with pinned memory:
```
[qwen-image-block] Weight swap timing at 100:
  find modules: 0.15ms
  synchronize before swap: 0.03ms
  cuda to cpu: 1.68ms
  wait cpu: 0.05ms
  cpu to cuda: 0.30ms
  synchronize: 32.17ms
Overall time: 35.19ms, total time in sections: 34.37ms
```

Therefore, I think this is probably due to the WDDM mode driver.

I think you can reduce the fluctuations in transfer speed, the transfer speed itself, and the time it takes to sync by not using other apps while training, turning off Hardware-Accelerated GPU Scheduling, etc.

FurkanGozukara:
@kohya-ss thank you so much it was as you said

i turned of hardware acceleration restarted pc and in 10 steps it is now like 50% faster :D 

17.39s to 11.73s/it, in 10 steps

[new_logs.txt](https://github.com/user-attachments/files/23263156/new_logs.txt)

kohya-ss:
> i turned of hardware acceleration restarted pc and in 10 steps it is now like 50% faster :D

That's great😀

FurkanGozukara:
> turning off Hardware-Accelerated GPU Scheduling,

thank you so much

are there any other optimization we can make other than this? "turning off Hardware-Accelerated GPU Scheduling,"

FurkanGozukara:
also can you merge it if possible? @kohya-ss ty so much

FurkanGozukara:
Ok I guess i talked too early

i am doing a longer test with FP8 Scaled and sadly we are still way way slower than Linux - I tested RTX 5090 on linux

what is the real culprit i really wonder

image

my BUS interface is 8x

maybe that is another culprit i need to check this :D 

image


when i train my second gpu rtx 3090 ti, its speed like almost half of rtx 5090 and it has good amount of watt usage

so bus interface is not the culprit - this one uses 4x only atm

image

kohya-ss:
It appears you have two GPUs, but both the screen display and training appear to be running on GPU 0. Connecting a display to GPU 1 may be slightly faster.

Gen 4 x8 has a bandwidth of approximately 15GB/s. Since 16/60 blocks are transferred in forward and backward pass, and also in both CPU->GPU and GPU->CPU directions, if the model weights are 20GB (fp8 scaled), a transfer of 20 * (16/60) * 2 * 2, approximately 20GB is required. Transferring 20GB will take about 2 seconds in Gen 4 x8, including overhead. Since one step takes 7-8 seconds, if the transfer is completely asynchronous, Gen 4 x8 should be sufficient.

FurkanGozukara:
@kohya-ss i just tested this

moved display cables to second gpu

restarted pc

did 40 steps without pinned memory

speed is 6.76 second / it

with pinned memory it was 6.47 second / it

on linux I am getting like 4 second / it :( on runpod

so indeed it brought some speed increase but this linux speed, the speed on runpod is so extremely more compared to windows what we get



image


this one below from linux - fp8 scaled is 4 second / it

probably i can get some more faster with lesser block swap when gpu is empty lets see how faster i can push but 4 second / it seems impossible

image

FurkanGozukara:
ok with only 7 block swap i am able to push up to 6.10 second / it

with overlocking it gets even better since there is still space but can't reach linux :(

image

JonkoXL:
> Add `--use_pinned_memory_for_block_swap` for each training script to enable pinned memory. Will work with Windows and Linux, but tested with Windows only.
> 
> Qwen-Image fine tuning is tested.

Awesome!
Will you add it for Wan as well? May be beneficial for block swapping with longer / higher res video training.

kohya-ss:
> Awesome!
> Will you add it for Wan as well? May be beneficial for block swapping with longer / higher res video training.

The option is added to Wan/HunyuanVideo/FramePack/Flux.1 Kontext too. However, unfortunately, for long videos and high-resolution videos, the calculation time is longer than the transfer time, so I think the improvement is limited.

kohya-ss:
In my environment, without fp8 scaled and 40 block swap is 4sec/it, so the WDDM mode driver is likely to be the cause.

To improve GPU performance in a Windows environment, it is generally a good idea to check the following:

- NVIDIA Control Panel, 3D settings 
  - V-Sync: Off
  - Power management mode: Prefer maximum performance
- Windows Power Plan: High Performance
- PCIe Link State Power Management: Off
  - https://www.reddit.com/r/witcher/comments/3at2st/pcie_power_management_can_throttle_your_gpu/
- Enable Resizable BAR (ReBAR) in BIOS settings

kohya-ss:
Speed ​​improvement of about 5% when not using pin memory by 797c38a.

FurkanGozukara:
I tried to turn off WDDM but sadly shameless nvidia blocked it for even RTX 5090

so only way is using native Linux i presume? yes Linux speed is 4 second / it for rtx 5090 i tested on runpod

I will check  Resizable BAR (ReBAR) in BIOS settings first time hearing

FurkanGozukara:
I enabled ReBAR but it didnt make big difference

I was expecting huge. 

By the way turning off hardware accelerated GPU scheduling indeed makes big difference for pinned memory

Maybe we can get some ideas from here : https://github.com/Orbmu2k/nvidiaProfileInspector/issues/357 

image

kohya-ss:
The latest implementation reuses pinned memory and use two CUDA streams (GPU->CPU and CPU->GPU). The latest implementation is faster on the RTX A6000, but the previous implementation seems to be faster on the RTX 6000 Blackwell...

kohya-ss:
One stream seems to be faster than two streams.

kohya-ss:
On RTX 6000 Blackwell Max-Q (Windows, TCC mode), training of Qwen-Image LoRA (128 rank, 1328x1328, batch size 1, fp8 scaled) can be achieved with almost no performance drop using both using pinned memory with block swap 40, and non-pinned with block swap 10. Power consumption also remains consistently above 290W.

Edit: On RTX A6000 (Windows,  WDDM mode), non-pinned is faster than pinned.

FurkanGozukara:
> TCC mode

is the key. shameless NVIDIA blocked it for RTX 5090 :( 

thank you so much for trying more

I will test latest config on Windows sadly in WDDM mode

FurkanGozukara:
looks like enabling TCC mode is not that hard on Windows

https://www.bilibili.com/opus/891652532297793543

i just requires a driver hack

i wonder if we could simplify that

kohya-ss:
LoRA training for Qwen-Image, FramePack, Wan, FLUX.1 Kontext and HunyuanVideo has been tested.

FurkanGozukara:
@kohya-ss can we add this for fine tuning as well please

ty

xzuyn:
comparison on 7900xtx. seems faster, but total time is slightly longer .

pinned memory disabled: https://gist.github.com/xzuyn/bcc2ba07673923cc3be047f203ed8bd8
pinned memory enabled: https://gist.github.com/xzuyn/9a3320f0917e75807377a5b77b9565ba

FurkanGozukara:
@xzuyn thanks for test. so If AMD was selling GPUs from 96 GB VRAM instead of same as NVIDIA they would make sense to purchase 

I can't understand how incompetent they are :(

PR #706, opened 2025-11-03
Add support for `pixi`

hameerabbasi:
This PR adds support for a `pixi` development environment setup.

`pixi` is often touted as a `conda` replacement, but it is much more; including workflow running and dev environment management.

For example, with this PR and `pixi` installed, one only needs to do `pixi run lint` in order to lint everything.

It also adds configurations for ROCm 6.4 and CUDA 13.0 (for newer GPUs).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ GEMINI.md
 .gemini/
 AGENTS.md
 .vscode/settings.json
+# pixi environments
+.pixi/*
+!.pixi/config.toml
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,0 +1,4331 @@
+version: 6
+environments:
+  cu124:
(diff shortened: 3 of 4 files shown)


kohya-ss:
Thank you for submitting this PR. I've looked into Pixi, and it's indeed an impressive tool with excellent features.

However, I have some concerns about introducing it to this project:

1. **Overlap with existing setup**: We currently manage our environments, including CUDA configurations, with uv/pip. The additional benefits of adding Pixi seem limited in our case.

2. **User impact**: Most existing users are familiar with uv/pip, so discontinuing them would have significant impact. On the other hand, maintaining both systems in parallel would increase the maintenance burden.

3. **Environment diversity**: Our users have diverse environments (RTX 30/40/50 series, various CUDA versions, etc.), so our current flexible dependency management (not pinning to specific versions) seems more suitable for accommodating this variety.

4. **Maintenance considerations**: Since I don't regularly use Linux as my development environment, it would be challenging for me to address Linux-specific issues properly.

If there are benefits to introducing Pixi that outweigh these concerns, I'd love to hear about them. Particularly, if there are problems that cannot be solved with the current uv/pip setup, I'd be happy to reconsider.

Also, if Pixi becomes more widely adopted in the future, I'd be open to revisiting this discussion.

PR #717, opened 2025-11-09
Profiling util

dxqb:
Hi kohya,

your code has helped me find a performance improvement for OneTrainer, so I thought I contribute back what I've used to benchmark.
It just integrates torch profiler to gain some performance insights like this

image


kohya-ss:
Thank you for this, and I'm glad my code was helpful :) This looks very useful. I'll give this code a try.

PR #722, opened 2025-11-13
feat: Add `torch.compile` support to training scripts.

kohya-ss:
Added `torch.compile` support with the following options:
- `--compile`: Enable torch.compile
- `--compile_backend`: Backend selection (default: `inductor`)
- `--compile_mode`: Compilation mode (default: `default`)
- `--compile_dynamic`: Enable dynamic shapes
- `--compile_fullgraph`: Enable fullgraph mode
- `--compile_cache_size_limit`: Cache size limit for torch.compile (optional)
  - Larger values (e.g., 32) may improve performance for models with varying input shapes
  - Increases memory usage

Note: The `--compile_args` format is deprecated and will be removed in future versions.

Training:
- [x] Qwen-Image/Edit/Edit-2509
- [x] Wan
- [x] FramePack
- [x] FLUX.1 Kontext
- [x] HunyuanVideo

Generation:
- [x] All generation scripts

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -58,6 +58,10 @@
 
 GitHub Discussionsを有効にしました。コミュニティのQ&A、知識共有、技術情報の交換などにご利用ください。バグ報告や機能リクエストにはIssuesを、質問や経験の共有にはDiscussionsをご利用ください。[Discussionはこちら](https://github.com/kohya-ss/musubi-tuner/discussions)
 
+- 2025/11/15
+    - 学習および推論スクリプトで `torch.compile` を使用できるようになりました。[PR #722](https://github.com/kohya-ss/musubi-tuner/pull/722)
+        - `--compile` オプションを指定すると、`torch.compile` を使用して学習および推論を行います。詳細は[ドキュメント](./docs/torch_compile.md)を参照してください。
+
 - 2025/11/02
     - 各学習スクリプトに `--use_pinned_memory_for_block_swap` オプションを追加しました。またblock swapの処理自体も改善しました。[PR #700](https://github.com/kohya-ss/musubi-tuner/pull/700)
         - このオプションを指定すると、block swapのoffloadingにピン留めメモリを使用します。これによりblock swapのパフォーマンスが向上する可能性があります。ただし、Windows環境の場合は共有GPUメモリの使用量が増加します。詳しくは[ドキュメント](./docs/hunyuan_video.md#memory-optimization)を参照してください。
@@ -72,14 +76,6 @@ GitHub Discussionsを有効にしました。コミュニティのQ&A、知識�
     - 制御画像を持つ画像データセットで、対象画像と制御画像の組み合わせが正しく読み込まれない不具合を修正しました。[PR #684](https://github.com/kohya-ss/musubi-tuner/pull/684)
         - **制御画像を持つ画像データセットをご利用の場合、latentキャッシュの再作成を行ってください。**
         - 先頭のマッチだけで判断していたため、対象画像が`a.png`、`ab.png`で、制御画像が`a_1.png`、`ab_1.png`のとき、`a.png`には、`a_1.png`と`ab_1.png`が組み合わされてしまっていました。
-
-- 2025/10/13
-    - Qwen-Image-Edit、2509の推論スクリプトで、ピクセル単位での生成画像の一貫性を向上するReference Consistency Mask (RCM)機能を追加しました。[PR #643]
-    (https://github.com/kohya-ss/musubi-tuner/pull/643)
-        - RCMは、生成画像が制御画像と比較してわずかな位置ずれを起こす問題を解決します。詳細は[Qwen-Imageのドキュメント](./docs/qwen_image.md#inpainting-and-reference-consistency-mask-rcm)を参照してください。
-    - あわせて同PRにて `--resize_control_to_image_size` オプションが指定されていない場合でも、コントロール画像が出力画像と同じサイズにリサイズされてしまう不具合を修正しました。**生成画像が変化する可能性がありますので、オプションを確認してください。** 
-    - FramePackの1フレーム推論で、`--one_frame_auto_resize`オプションを追加しました。[PR #646](https://github.com/kohya-ss/musubi-tuner/pull/646)
-        - 生成画像の解像度を自動的に調整します。`--one_frame_inference`を指定した場合にのみ有効です。詳細は[FramePackの1フレーム推論のドキュメント](./docs/framepack_1f.md#one-single-frame-inference--1フレーム推論)を参照してください。
 
 ### リリースについて
 
@@ -144,6 +140,7 @@ Musubi Tunerの解説記事執筆や、関連ツールの開発に取り組ん�
 - [高度な設定](./docs/advanced_config.md)
 - [学習中のサンプル生成](./docs/sampling_during_training.md)
 - [ツールとユーティリティ](./docs/tools.md)
+- [torch.compileの使用方法](./docs/torch_compile.md)
 
 ## インストール
 
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ If you find this project helpful, please consider supporting its development via
 
 GitHub Discussions Enabled: We've enabled GitHub Discussions for community Q&A, knowledge sharing, and technical information exchange. Please use Issues for bug reports and feature requests, and Discussions for questions and sharing experiences. [Join the conversation →](https://github.com/kohya-ss/musubi-tuner/discussions)
 
+- November 15, 2025
+    - Added support for `torch.compile` in training and inference scripts. See [PR #722](https://github.com/kohya-ss/musubi-tuner/pull/722).
+        - By specifying the `--compile` option, training and inference will utilize `torch.compile`. Please refer to the [documentation](./docs/torch_compile.md) for details.
+
 - November 2, 2025
     - Added `--use_pinned_memory_for_block_swap` option to each training script and improved the block swap process itself. See [PR #700](https://github.com/kohya-ss/musubi-tuner/pull/700).
         - When specified, this option uses pinned memory for block swap offloading. This may improve block swap performance. However, on Windows environments, it increases shared GPU memory usage. Please refer to the [documentation](./docs/hunyuan_video.md#memory-optimization) for details.
@@ -77,13 +81,6 @@ GitHub Discussions Enabled: We've enabled GitHub Discussions for community Q&A, 
     - Fixed a bug in image datasets with control images where the combination of target and control images was not loaded correctly. See [PR #684](https://github.com/kohya-ss/musubi-tuner/pull/684).
         - **If you are using an image dataset with control images, please recreate the latent cache.**
         - Since only the first match was used for judgment, when the target images were `a.png` and `ab.png`, and the control images were `a_1.png` and `ab_1.png`, both `a_1.png` and `ab_1.png` were combined with `a.png`.
-
-- October 13, 2025
-    - Added Reference Consistency Mask (RCM) feature to Qwen-Image-Edit, 2509 inference script to improve pixel-level consistency of generated images. See [PR #643](https://github.com/kohya-ss/musubi-tuner/pull/643)
-        - RCM addresses the issue of slight positional drift in generated images compared to the control image. For details, refer to the [Qwen-Image documentation](./docs/qwen_image.md#inpainting-and-reference-consistency-mask-rcm).
-    - Fixed a bug where the control image was being resized to match the output image size even when the `--resize_control_to_image_size` option was not specified. **This may change the generated images, so please check your options.**
-    - FramePack 1-frame inference now includes the `--one_frame_auto_resize` option. [PR #646](https://github.com/kohya-ss/musubi-tuner/pull/646)
-        - Automatically adjusts the resolution of the generated image. This option is only effective when `--one_frame_inference` is specified. For details, refer to the [FramePack 1-frame inference documentation](./docs/framepack_1f.md#one-single-frame-inference--1フレーム推論).
 
 ### Releases
 
@@ -148,6 +145,7 @@ For detailed information on specific architectures, configurations, and advanced
 - [Advanced Configuration](./docs/advanced_config.md)
 - [Sampling during Training](./docs/sampling_during_training.md)
 - [Tools and Utilities](./docs/tools.md)
+- [Using torch.compile](./docs/torch_compile.md)
 
 ## Installation
 


FurkanGozukara:
Amazing thank you so much

sdbds:
Therefore, should we also abolish the compile interface of dynamo_backend and accelerate, and directly use the native version of PyTorch?

sdbds:
I reviewed the code and it looks good. Block swap is also supported by disabling the compilation of linear layers.
I remember that back when we used Accelerate to wrap the model (back in ControlNet), if we added compilation, we needed to re-wrap the original part.
It's uncertain whether we need to do this during fine-tuning.

TingTingin:
any speed tests on windows?

kohya-ss:
> Therefore, should we also abolish the compile interface of dynamo_backend and accelerate, and directly use the native version of PyTorch?

I personally don't use dynamo_backend, but if we don't use the `--compile` option, the behavior will be the same as before, so I think it's fine to leave it as it is.

> I remember that back when we used Accelerate to wrap the model (back in ControlNet), if we added compilation, we needed to re-wrap the original part.

I'm not sure, but the current implementation compiles the model before wrapping with accelerate, so it might work as is.

kohya-ss:
## Quick Benchmark Results

Training time for epochs 2-5 when training Qwen-Image at 1328×1328 resolution with batch size=1, rank=4 LoRA for 5 epochs.

**Note:** Epoch 1 takes significantly longer when `dynamic=False` due to compilation overhead (*1).

### RTX A6000 (Power Limit: 180W)

| Configuration | Time (mm:ss) | Time (seconds) | Speedup |
|---------------|--------------|----------------|---------|
| No compile | 7:52 | 472s | 1.00x (baseline) |
| `mode=default` | 7:07 | 427s | 1.11x (~10.5% faster) |
| `mode=max-autotune-no-cudagraphs` | 7:05 | 425s | 1.11x (~11.1% faster) |
| `mode=default, dynamic` | 7:17 | 437s | 1.08x (~8.0% faster) |
| `mode=max-autotune-no-cudagraphs, dynamic` | 7:14 | 434s | 1.09x (~8.8% faster) |

### RTX PRO 6000 Blackwell Max-Q Workstation Edition (Power Limit: 250W)

| Configuration | Time (mm:ss) | Time (seconds) | Speedup |
|---------------|--------------|----------------|---------|
| No compile | 2:44 | 164s | 1.00x (baseline) |
| `mode=default` | 2:18 | 138s | 1.19x (~18.8% faster) |
| `mode=max-autotune-no-cudagraphs` | 2:11 | 131s | 1.25x (~25.2% faster) |
| `mode=default, dynamic` | 2:16 | 136s | 1.21x (~20.6% faster) |
| `mode=max-autotune-no-cudagraphs, dynamic` | 2:11 | 131s | 1.25x (~25.2% faster) |
| `mode=max-autotune-no-cudagraphs, dynamic, allow_tf32=True` | 2:10 | 130s | 1.26x (~26.2% faster) |

*1 Musubi Tuner widely uses `--split_attn`, which speeds up attention by not using attention masks, so the input shape is different each time. Therefore, `dynamic=False` causes many compilations. 
If the batch size is 1, all patterns of all lengths are covered in the first epoch. However, if the batch size is 2 or more, the number of combinations becomes very large. Therefore, `dynamic=False` is only recommended when the batch size is 1.

sdbds:
> > Therefore, should we also abolish the compile interface of dynamo_backend and accelerate, and directly use the native version of PyTorch?
> 
> I personally don't use dynamo_backend, but if we don't use the `--compile` option, the behavior will be the same as before, so I think it's fine to leave it as it is.
> 
> > I remember that back when we used Accelerate to wrap the model (back in ControlNet), if we added compilation, we needed to re-wrap the original part.
> 
> I'm not sure, but the current implementation compiles the model before wrapping with accelerate, so it might work as is.

```
Traceback (most recent call last):
  File "E:\Code\musubi-tuner-scripts\musubi-tuner\qwen_image_train_network.py", line 4, in <module>
    main()
  File "E:\Code\musubi-tuner-scripts\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 521, in main
    trainer.train(args)
  File "E:\Code\musubi-tuner-scripts\musubi-tuner\src\musubi_tuner\hv_train_network.py", line 1895, in train
    accelerator.unwrap_model(transformer).move_to_device_except_swap_blocks(accelerator.device)  # reduce peak memory usage
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\accelerator.py", line 2866, in unwrap_model
    return extract_model_from_parallel(model, keep_fp32_wrapper, keep_torch_compile)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\utils\other.py", line 180, in extract_model_from_parallel
    model = model._orig_mod
            ^^^^^^^^^^^^^^^
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\torch\nn\modules\module.py", line 1964, in __getattr__
    raise AttributeError(
AttributeError: 'QwenImageTransformer2DModel' object has no attribute '_orig_mod'
```

Just as I suspected, the accelerate.unwrap_model  part does indeed need to be re-encapsulated as orig_mod...

test params
```
  --text_encoder=./ckpts/text_encoder/qwen_2.5_vl_7b.safetensors
  --edit_plus
  --flash_attn
  --split_attn
  --timestep_sampling=qwen_shift
  --guidance_scale=1
  --max_train_epochs=40
  --gradient_checkpointing
  --gradient_accumulation_steps=1
  --network_args
  loraplus_lr_ratio=4
  --network_dim=32
  --network_alpha=16
  --network_module=networks.lora_qwen_image
  --gradient_accumulation_steps=1
  --lr_scheduler=cosine_with_min_lr
  --lr_scheduler_num_cycles=1
  --lr_decay_steps=0.2
  --lr_scheduler_min_lr_ratio=0.1
  --mixed_precision=bf16
  --fp8_base
  --fp8_scaled
  --persistent_data_loader_workers
  --blocks_to_swap=16
  --compile
  --compile_backend=inductor
  --compile_mode=reduce-overhead
  --compile_fullgraph
  --compile_dynamic
  --img_in_txt_in_offloading
  --optimizer_type=adv_optm.AdamW_adv
  --optimizer_args
  use_atan2=True
  grams_moment=True
  nnmf_factor=True
  --save_every_n_epochs=2
  --sample_at_first
  --sample_every_n_epochs=2
  --sample_prompts=./toml/qinglong_qwen_image_edit.txt
```

kohya-ss:
If you get the error `RuntimeError: Compiler: cl is not found.` when `dynamic=True`, you may need to run the script in `x64 Native Tools Command Prompt for VS 2022`.

sdbds:
According to accelerator's problem,  `accelerator.unwrap_model(transformer,keep_torch_compile=True)`

FurkanGozukara:
Tested on Windows 11, RTX 5090, Qwen Image  LoRA FP8 Scaled, LoRA Rank 64, BS1, 1328x1328

For testing I had to modify time calculation and start average time after 1st step done. Please make this fix @kohya-ss . otherwise initial overhead making impossible to get true average speed - you have to wait too many steps. 

So average speed must be calculated starting from second step and so on. Please make this change both on Musubi and SD Scripts. 

Tested for 56 steps - 1 epoch was 28 images

I didn't use block swap for these tests.  

when just used  `--compile`

5.39 second / it > 4.65 second / it : 13.72% reduced time 

When used like

`--compile --compile_mode max-autotune-no-cudagraphs`

5.39 second / it > 4.50 second / it : 16.51% reduced time 

max-autotune-no-cudagraphs didnt make big improvement but it brings huge overhead before starting. however second training didn't recompile - it was immediate

`--compile_dynamic` fails as below

```
  total optimization steps / 学習ステップ数: 5600
INFO:musubi_tuner.hv_train_network:set DiT model name for metadata: Q:/SECourses_Musubi_Trainer_v19/Training_Models_Qwen/qwen_image_bf16.safetensors
INFO:musubi_tuner.hv_train_network:set VAE model name for metadata: Q:/SECourses_Musubi_Trainer_v19/Training_Models_Qwen/qwen_train_vae.safetensors
steps:   0%|                                                                                                                                                | 0/5600 [00:00<?, ]INFO:musubi_tuner.hv_train_network:DiT dtype: torch.bfloat16, device: cuda:0

epoch 1/200
Trying to import sageattention
Successfully imported sageattention
Trying to import sageattention
Successfully imported sageattention
INFO:musubi_tuner.dataset.image_video_dataset:epoch is incremented. current_epoch: 0, epoch: 1
INFO:musubi_tuner.dataset.image_video_dataset:epoch is incremented. current_epoch: 0, epoch: 1
F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_decomp\decompositions.py:1462: UserWarning: record_context_cpp is not support on non-linux non-x86_64 platforms (Triggered internally at C:\actions-runner\_work\pytorch\pytorch\pytorch\torch\csrc\profiler\unwind\unwind.cpp:12.)
  out = alpha * torch.mm(mat1, mat2)
Traceback (most recent call last):
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 525, in <module>
    main()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 521, in main
    trainer.train(args)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\hv_train_network.py", line 2182, in train
    model_pred, target = self.call_dit(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 468, in call_dit
    model_pred = model(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1773, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1784, in _call_impl
    return forward_call(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\utils\operations.py", line 814, in forward
    return model_forward(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\utils\operations.py", line 802, in __call__
    return convert_to_fp32(self.model_forward(*args, **kwargs))
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\amp\autocast_mode.py", line 44, in decorate_autocast
    return func(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image\qwen_image_model.py", line 1176, in forward
    encoder_hidden_states, hidden_states = self._gradient_checkpointing_func(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image\qwen_image_model.py", line 1096, in _gradient_checkpointing_func
    return torch.utils.checkpoint.checkpoint(block, *args, use_reentrant=False)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_compile.py", line 53, in inner
    return disable_fn(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_dynamo\eval_frame.py", line 929, in _fn
    return fn(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\utils\checkpoint.py", line 495, in checkpoint
    ret = function(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_dynamo\eval_frame.py", line 375, in __call__
    return super().__call__(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1773, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1784, in _call_impl
    return forward_call(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_dynamo\eval_frame.py", line 749, in compile_wrapper
    raise e.remove_dynamo_frames() from None  # see TORCHDYNAMO_VERBOSE=1
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\compile_fx.py", line 923, in _compile_fx_inner
    raise InductorError(e, currentframe()).with_traceback(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\compile_fx.py", line 907, in _compile_fx_inner
    mb_compiled_graph = fx_codegen_and_compile(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\compile_fx.py", line 1578, in fx_codegen_and_compile
    return scheme.codegen_and_compile(gm, example_inputs, inputs_to_check, graph_kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\compile_fx.py", line 1456, in codegen_and_compile
    compiled_module = graph.compile_to_module()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\graph.py", line 2293, in compile_to_module
    return self._compile_to_module()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\graph.py", line 2303, in _compile_to_module
    mod = self._compile_to_module_lines(wrapper_code)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\graph.py", line 2371, in _compile_to_module_lines
    mod = PyCodeCache.load_by_key_path(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\codecache.py", line 3296, in load_by_key_path
    mod = _reload_python_module(key, path, set_sys_modules=in_toplevel)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\runtime\compile_tasks.py", line 31, in _reload_python_module
    exec(code, mod.__dict__, mod.__dict__)
  File "C:\Users\Furkan\AppData\Local\Temp\torchinductor_Furkan\sv\csvoukxpzeeu722fceaob34cwhaizegi2b6zmny463bse6jqzky7.py", line 190, in <module>
    cpp_fused_4 = async_compile.cpp_pybinding(['const double*', 'const double*', 'float*', 'float*'], '''
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\async_compile.py", line 434, in cpp_pybinding
    return CppPythonBindingsCodeCache.load_pybinding(argtypes, source_code)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\codecache.py", line 2783, in load_pybinding
    return cls.load_pybinding_async(*args, **kwargs)()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\codecache.py", line 2775, in future
    result = get_result()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\codecache.py", line 2572, in load_fn
    result = worker_fn()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\codecache.py", line 2601, in _worker_compile_cpp
    builder.build()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\cpp_builder.py", line 1718, in build
    run_compile_cmd(build_cmd, cwd=_build_tmp_dir)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\cpp_builder.py", line 401, in run_compile_cmd
    _run_compile_cmd(cmd_line, cwd)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_inductor\cpp_builder.py", line 396, in _run_compile_cmd
    raise exc.CppCompileError(cmd, output) from e
torch._inductor.exc.InductorError: CppCompileError: C++ compile error

Command:
cl /I C:/Python310/Include /I F:/SECourses_Musubi_Trainer_v16/SECourses_Musubi_Trainer/venv/lib/site-packages/torch/include /I F:/SECourses_Musubi_Trainer_v16/SECourses_Musubi_Trainer/venv/lib/site-packages/torch/include/torch/csrc/api/include /D TORCH_INDUCTOR_CPP_WRAPPER /D STANDALONE_TORCH_HEADER /D C10_USING_CUSTOM_GENERATED_MACROS /DLL /MD /O2 /std:c++20 /wd4819 /wd4251 /wd4244 /wd4267 /wd4275 /wd4018 /wd4190 /wd4624 /wd4067 /wd4068 /EHsc /openmp /openmp:experimental C:/Users/Furkan/AppData/Local/Temp/torchinductor_Furkan/4g/c4gvnqradrvpf7q5zbrwpxewnejv2onqyqxu7kcnfh4d5xom3bns.main.cpp /FeC:/Users/Furkan/AppData/Local/Temp/torchinductor_Furkan/4g/c4gvnqradrvpf7q5zbrwpxewnejv2onqyqxu7kcnfh4d5xom3bns.main.pyd /LD /link /LIBPATH:C:/Python310/libs /LIBPATH:F:/SECourses_Musubi_Trainer_v16/SECourses_Musubi_Trainer/venv/lib/site-packages/torch/lib torch.lib torch_cpu.lib torch_python.lib sleef.lib

Output:
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35215 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

cl : Command line warning D9025 : overriding '/openmp' with '/openmp:experimental'
c4gvnqradrvpf7q5zbrwpxewnejv2onqyqxu7kcnfh4d5xom3bns.main.cpp
F:/SECourses_Musubi_Trainer_v16/SECourses_Musubi_Trainer/venv/lib/site-packages/torch/include\torch/csrc/inductor/cpp_prefix.h(3): fatal error C1083: Cannot open include file: 'omp.h': No such file or directory


Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

steps:   0%|                                                                                                                                                | 0/5600 [00:06<?, ]
Traceback (most recent call last):
  File "C:\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\Scripts\accelerate.exe\__main__.py", line 6, in <module>
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\commands\accelerate_cli.py", line 50, in main
    args.func(args)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\commands\launch.py", line 1213, in launch_command
    simple_launcher(args)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\commands\launch.py", line 795, in simple_launcher
    raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
subprocess.CalledProcessError: Command '['F:\\SECourses_Musubi_Trainer_v16\\SECourses_Musubi_Trainer\\venv\\Scripts\\python.exe', 'F:/SECourses_Musubi_Trainer_v16/SECourses_Musubi_Trainer/musubi-tuner/src/musubi_tuner/qwen_image_train_network.py', '--config_file', 'F:/SECourses_Musubi_Trainer_v16/test_trainings\\My_LoRA_Training_20251113-135616.toml', '--compile', '--compile_dynamic']' returned non-zero exit status 1.
13:57:00-602396 ERROR    Training failed with exit code: 1
```


`--compile_fullgraph` fails as below


```
  batch size per device / バッチサイズ: 1
  gradient accumulation steps / 勾配を合計するステップ数 = 1
  total optimization steps / 学習ステップ数: 5600
INFO:musubi_tuner.hv_train_network:set DiT model name for metadata: Q:/SECourses_Musubi_Trainer_v19/Training_Models_Qwen/qwen_image_bf16.safetensors
INFO:musubi_tuner.hv_train_network:set VAE model name for metadata: Q:/SECourses_Musubi_Trainer_v19/Training_Models_Qwen/qwen_train_vae.safetensors
steps:   0%|                                                                                                                                                | 0/5600 [00:00<?, ]INFO:musubi_tuner.hv_train_network:DiT dtype: torch.bfloat16, device: cuda:0

epoch 1/200
Trying to import sageattention
Successfully imported sageattention
Trying to import sageattention
Successfully imported sageattention
INFO:musubi_tuner.dataset.image_video_dataset:epoch is incremented. current_epoch: 0, epoch: 1
INFO:musubi_tuner.dataset.image_video_dataset:epoch is incremented. current_epoch: 0, epoch: 1
Traceback (most recent call last):
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 525, in <module>
    main()
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 521, in main
    trainer.train(args)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\hv_train_network.py", line 2182, in train
    model_pred, target = self.call_dit(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 468, in call_dit
    model_pred = model(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1773, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1784, in _call_impl
    return forward_call(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\utils\operations.py", line 814, in forward
    return model_forward(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\utils\operations.py", line 802, in __call__
    return convert_to_fp32(self.model_forward(*args, **kwargs))
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\amp\autocast_mode.py", line 44, in decorate_autocast
    return func(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image\qwen_image_model.py", line 1176, in forward
    encoder_hidden_states, hidden_states = self._gradient_checkpointing_func(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image\qwen_image_model.py", line 1096, in _gradient_checkpointing_func
    return torch.utils.checkpoint.checkpoint(block, *args, use_reentrant=False)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_compile.py", line 53, in inner
    return disable_fn(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_dynamo\eval_frame.py", line 929, in _fn
    return fn(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\utils\checkpoint.py", line 495, in checkpoint
    ret = function(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_dynamo\eval_frame.py", line 375, in __call__
    return super().__call__(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1773, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\nn\modules\module.py", line 1784, in _call_impl
    return forward_call(*args, **kwargs)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\torch\_dynamo\eval_frame.py", line 745, in compile_wrapper
    raise e.with_traceback(None) from e.__cause__  # User compiler error
torch._dynamo.exc.Unsupported: Skip calling `torch.compiler.disable()`d function
  Explanation: Skip calling function `<function apply_rotary_emb_qwen at 0x00000190FE558790>` since it was wrapped with `torch.compiler.disable` (reason: None)
  Hint: Remove the `torch.compiler.disable` call

  Developer debug context: <function apply_rotary_emb_qwen at 0x00000190FE558790>


from user code:
   File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image\qwen_image_model.py", line 900, in forward
    attn_output = self.attn(
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\musubi-tuner\src\musubi_tuner\qwen_image\qwen_image_model.py", line 716, in forward
    img_query = apply_rotary_emb_qwen(img_query, img_freqs, use_real=False)

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

steps:   0%|                                                                                                                                                | 0/5600 [00:03<?, ]
Traceback (most recent call last):
  File "C:\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\Scripts\accelerate.exe\__main__.py", line 6, in <module>
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\commands\accelerate_cli.py", line 50, in main
    args.func(args)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\commands\launch.py", line 1213, in launch_command
    simple_launcher(args)
  File "F:\SECourses_Musubi_Trainer_v16\SECourses_Musubi_Trainer\venv\lib\site-packages\accelerate\commands\launch.py", line 795, in simple_launcher
    raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
subprocess.CalledProcessError: Command '['F:\\SECourses_Musubi_Trainer_v16\\SECourses_Musubi_Trainer\\venv\\Scripts\\python.exe', 'F:/SECourses_Musubi_Trainer_v16/SECourses_Musubi_Trainer/musubi-tuner/src/musubi_tuner/qwen_image_train_network.py', '--config_file', 'F:/SECourses_Musubi_Trainer_v16/test_trainings\\My_LoRA_Training_20251113-135800.toml', '--compile', '--compile_fullgraph']' returned non-zero exit status 1.
13:58:44-467134 ERROR    Training failed with exit code: 1
13:58:49-079404 INFO     Train Qwen Image model...
13:58:49-081674 INFO     Dataset config saved to: F:/SECourses_Musubi_Trainer_v16/test_trainings\2025_11_13_13_58_49_080.toml
```

kohya-ss:
Thank you for checking!

> According to accelerator's problem, `accelerator.unwrap_model(transformer,keep_torch_compile=True)`

Hmm, it looks like the default value for `keep_torch_compile` is True... 

I tried DDP distributed training in my environment and it worked. However, it seems that `torch.compile` can be done after `accelerate.prepare`, so I will move it there.

kohya-ss:
> For testing I had to modify time calculation and start average time after 1st step done. Please make this fix @kohya-ss . otherwise initial overhead making impossible to get true average speed - you have to wait too many steps.

In Musubi Tuner, the input shape is different each step, so compilation is required not only for the first step but also for subsequent steps. Therefore, I don't think it's appropriate to exclude the first step just for measurement. Please measure after compilation is complete.

> `--compile_dynamic` fails as below

Please run the script in `x64 Native Tools Command Prompt for VS 2022`. This means you need Visual Studio 2022 or the Build Tools.

> `--compile_fullgraph` fails as below

Fullgraph doesn't work yet for a few reasons. First, as an error occurs, complex operations need to be included in the compilation (it seems faster to exclude complex operations from compilation when `fullgraph=False`). Also, `--split_attn` no longer works with fullgraph, so we need to either use attention mask in SDPA (which slows down the process and eliminates the benefits of compilation) or set cu_seq_lens in FlashAttention (this is not yet implemented). Please wait until cu_seq_lens in FlashAttention is properly implemented. I think it will be a separate pull request.

kohya-ss:
`--compile_cache_size_limit` option is added. `32` is recommended. Please specify this option if you encounter the error about cache size.

sdbds:
You can obtain the VS tool development environment by running a script:
```
if (-not (Get-Command cl -ErrorAction SilentlyContinue)) {
  $vswhere = Join-Path ${env:ProgramFiles(x86)} "Microsoft Visual Studio\Installer\vswhere.exe"
  $vsPath = & $vswhere -latest -products * `
    -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 `
    -property installationPath
  & (Join-Path $vsPath "Common7\Tools\Launch-VsDevShell.ps1") -Arch amd64
}
```

kohya-ss:
All training and inference scripts seem to work with `--compile`. Other individual options have not been tested.

sdbds:
I retested it and it worked fine this time. The only problem seems to be that the block swap isn't working properly; it's too slow. I'm not sure if it's a compilation issue or a block swapping issue.
image
image

I'll try to wait...

kohya-ss:
> I retested it and it worked fine this time. The only problem seems to be that the block swap isn't working properly; it's too slow. I'm not sure if it's a compilation issue or a block swapping issue.

Thank you for confirming that it works.

If you have time, I would appreciate it if you could test it without `--compile` option and just use `--blocks_to_swap`.

sdbds:
> > I retested it and it worked fine this time. The only problem seems to be that the block swap isn't working properly; it's too slow. I'm not sure if it's a compilation issue or a block swapping issue.
> 
> Thank you for confirming that it works.
> 
> If you have time, I would appreciate it if you could test it without `--compile` option and just use `--blocks_to_swap`.

It appears to be a block swapping issue, possibly caused by the latest block swapping PR.
I set the block swap to 16, but it looks like only some are being used.
image

 ```
 --text_encoder=./ckpts/text_encoder/qwen_2.5_vl_7b.safetensors
  --edit_plus
  --flash_attn
  --split_attn
  --timestep_sampling=qwen_shift
  --guidance_scale=1
  --max_train_epochs=40
  --gradient_checkpointing
  --network_args
  loraplus_lr_ratio=4
  --network_dim=32
  --network_alpha=16
  --network_module=networks.lora_qwen_image
  --lr_scheduler=cosine_with_min_lr
  --lr_scheduler_num_cycles=1
  --lr_decay_steps=0.2
  --lr_scheduler_min_lr_ratio=0.1
  --mixed_precision=bf16
  --fp8_base
  --fp8_scaled
  --persistent_data_loader_workers
  --blocks_to_swap=16
  --img_in_txt_in_offloading
  --optimizer_type=adv_optm.AdamW_adv
  --optimizer_args
  use_atan2=True
  grams_moment=True
  nnmf_factor=True
  --save_every_n_epochs=2
```

Edit: I tried the Adamw optimizer, but the result was the same.

**Comment by kohya-ss** [comment]
> It appears to be a block swapping issue, possibly caused by the latest block swapping PR.
> I set the block swap to 16, but it looks like only some are being used.

When training with an image resolution of 1024x1024, one control image, batch size of 1, dim=32, AdamW optimizer, blocks_to_swap=16, and fp8_scaled, it consumed about 26GB of VRAM in my environment. Therefore, it may be that the slowness is due to the overflow of the dedicated VRAM.

**Comment by sdbds** [comment]
> > It appears to be a block swapping issue, possibly caused by the latest block swapping PR.
> > I set the block swap to 16, but it looks like only some are being used.
> 
> When training with an image resolution of 1024x1024, one control image, batch size of 1, dim=32, AdamW optimizer, blocks_to_swap=16, and fp8_scaled, it consumed about 26GB of VRAM in my environment. Therefore, it may be that the slowness is due to the overflow of the dedicated VRAM.

The new version of block swapping seems to be a bit different. Previously, if swap was set to a large value, shared video memory would be used more (blocks * 0.5G), while dedicated video memory would not overflow...
Because Windows NVIDIA drivers have an ECC mechanism, they consume an additional 0.5~1.5GB of video memory. Newer versions of the driver will cause block swapping to overflow...

**Comment by sdbds** [comment]
Additionally, if the sample function is enabled, it seems that the accelerator will still be used first...

**Comment by FurkanGozukara** [comment]
> > For testing I had to modify time calculation and start average time after 1st step done. Please make this fix @kohya-ss . otherwise initial overhead making impossible to get true average speed - you have to wait too many steps.
> 
> In Musubi Tuner, the input shape is different each step, so compilation is required not only for the first step but also for subsequent steps. Therefore, I don't think it's appropriate to exclude the first step just for measurement. Please measure after compilation is complete.
> 
> > `--compile_dynamic` fails as below
> 
> Please run the script in `x64 Native Tools Command Prompt for VS 2022`. This means you need Visual Studio 2022 or the Build Tools.
> 
> > `--compile_fullgraph` fails as below
> 
> Fullgraph doesn't work yet for a few reasons. First, as an error occurs, complex operations need to be included in the compilation (it seems faster to exclude complex operations from compilation when `fullgraph=False`). Also, `--split_attn` no longer works with fullgraph, so we need to either use attention mask in SDPA (which slows down the process and eliminates the benefits of compilation) or set cu_seq_lens in FlashAttention (this is not yet implemented). Please wait until cu_seq_lens in FlashAttention is properly implemented. I think it will be a separate pull request.


@kohya-ss you didnt understand what i mean 

when i start training it starts timing from the very first second

however if you start average step timing after 1st step done, you get accurate speed and timing - like estimated timing of training

so for example on runpod it starts with 20 second it for all images 1328x1328 gets down to 3.8 second / it which is the actual step speed since first step but since the initial loading etc takes so much time it shows as 20 second / it and you have to wait 100s of steps to see actual step speed

do you see this dramatic speed difference?

image


**it is because your timing starts from beginning and not accurate step speed until like 100s of steps done and it averages out. so start the timing after 1st step is done.** 


Also I have visual studio tools fully installed what else i have to do?

image

**Comment by FurkanGozukara** [comment]
by the way block swaps also got speed boost at me it works just fine 

i tested 20 block swap on qwen image 1328x1328, bs1

**Comment by sdbds** [comment]
> > It appears to be a block swapping issue, possibly caused by the latest block swapping PR.
> > I set the block swap to 16, but it looks like only some are being used.
> 
> When training with an image resolution of 1024x1024, one control image, batch size of 1, dim=32, AdamW optimizer, blocks_to_swap=16, and fp8_scaled, it consumed about 26GB of VRAM in my environment. Therefore, it may be that the slowness is due to the overflow of the dedicated VRAM.

Okay, everything's working now... It worked after I increased the block swap to 26...
It's strange that it only overflowed by about 4GB, yet it's using so many blocks...

**Comment by sdbds** [comment]
I think I've found the reason. It seems that more video memory is used during the initial compilation, and the video memory usage decreases after the compilation is complete.
I tested the 32GB 5090 and it had the same problem. It initially exceeded 32GB of video memory, but after using one block, it became 29.5GB.
Should we add an option to wait for compilation to complete before starting training, instead of doing it synchronously?

**Comment by kohya-ss** [comment]
> **it is because your timing starts from beginning and not accurate step speed until like 100s of steps done and it averages out. so start the timing after 1st step is done.**

Thank you for the explanation.

It's true that the first step takes a particularly long time. However, in reality, the entire first epoch often takes longer than the second and subsequent epochs. Therefore, to know the true speed, I recommend checking the speed from the second epoch onwards.
However, I think excluding the first step is an idea, so I would like to consider it in a separate PR.

> Also I have visual studio tools fully installed what else i have to do?

Activate the venv in that window and run the training script, and it should run with dynamic=True.

**Comment by kohya-ss** [comment]
> I think I've found the reason. It seems that more video memory is used during the initial compilation, and the video memory usage decreases after the compilation is complete.
> I tested the 32GB 5090 and it had the same problem. It initially exceeded 32GB of video memory, but after using one block, it became 29.5GB.

Thank you for the detailed verification. That makes a lot of sense. 

> Should we add an option to wait for compilation to complete before starting training, instead of doing it synchronously?

I think we'll run out of VRAM already on the first compilation... 

The results of the compilation are cached, so I think it would be more realistic to adjust blocks_to_swap and perform dummy training for just one epoch, and then perform actual training.

**Comment by FurkanGozukara** [comment]
> Activate the venv in that window and run the training script, and it should run with dynamic=True.

can you explain this? in my venv window when i type cl i still see it

image


also please add my recommendation it can be args

because this is causing huge issue to other people as well. they see first step speed which is sometimes being 5x or 10x of actual speed and it discourages them. i will just add args to configs so they will see actual step speed starting from step 2

usually people trains with batch size 1 and all images are same resolution so step speed is being constant

**Comment by sdbds** [comment]
> > I think I've found the reason. It seems that more video memory is used during the initial compilation, and the video memory usage decreases after the compilation is complete.
> > I tested the 32GB 5090 and it had the same problem. It initially exceeded 32GB of video memory, but after using one block, it became 29.5GB.
> 
> Thank you for the detailed verification. That makes a lot of sense.
> 
> > Should we add an option to wait for compilation to complete before starting training, instead of doing it synchronously?
> 
> I think we'll run out of VRAM already on the first compilation...
> 
> The results of the compilation are cached, so I think it would be more realistic to adjust blocks_to_swap and perform dummy training for just one epoch, and then perform actual training.

Most of the problems have been solved now. The only remaining issue is that using `sample at first` still results in an `orig_mod` error because enabling the first step of sampling still uses `accelerator.unwarpmodel`.

**Comment by kohya-ss** [comment]
> > Activate the venv in that window and run the training script, and it should run with dynamic=True.
> 
> can you explain this? in my venv window when i type cl i still see it

You can run `accelerate launch ...` command here.

If you use some script to run the training script, I think you can use the method sdbds wrote here: https://github.com/kohya-ss/musubi-tuner/pull/722#issuecomment-3527644341

**Comment by kohya-ss** [comment]
> Most of the problems have been solved now. The only remaining issue is that using `sample at first` still results in an `orig_mod` error because enabling the first step of sampling still uses `accelerator.unwarpmodel`.

In my environment, I don't get any errors when I compile with DDP and specify sample_at_first. Could you please share the stack trace?

**Comment by sdbds** [comment]
> > Most of the problems have been solved now. The only remaining issue is that using `sample at first` still results in an `orig_mod` error because enabling the first step of sampling still uses `accelerator.unwarpmodel`.
> 
> In my environment, I don't get any errors when I compile with DDP and specify sample_at_first. Could you please share the stack trace?

```
Traceback (most recent call last):
  File "E:\Code\musubi-tuner-scripts\musubi-tuner\qwen_image_train_network.py", line 4, in <module>
    main()
  File "E:\Code\musubi-tuner-scripts\musubi-tuner\src\musubi_tuner\qwen_image_train_network.py", line 522, in main
    trainer.train(args)
  File "E:\Code\musubi-tuner-scripts\musubi-tuner\src\musubi_tuner\hv_train_network.py", line 2146, in train
    self.sample_images(accelerator, args, 0, global_step, vae, transformer, sample_parameters, dit_dtype)
  File "E:\Code\musubi-tuner-scripts\musubi-tuner\src\musubi_tuner\hv_train_network.py", line 1099, in sample_images
    transformer = accelerator.unwrap_model(transformer)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\accelerator.py", line 2866, in unwrap_model
    return extract_model_from_parallel(model, keep_fp32_wrapper, keep_torch_compile)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\utils\other.py", line 180, in extract_model_from_parallel
    model = model._orig_mod
            ^^^^^^^^^^^^^^^
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\torch\nn\modules\module.py", line 1964, in __getattr__
    raise AttributeError(
AttributeError: 'QwenImageTransformer2DModel' object has no attribute '_orig_mod'
steps:   0%|                                                                                  | 0/4000 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\commands\launch.py", line 1208, in <module>
    main()
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\commands\launch.py", line 1204, in main
    launch_command(args)
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\commands\launch.py", line 1198, in launch_command
    simple_launcher(args)
  File "E:\Code\musubi-tuner-scripts\.venv\Lib\site-packages\accelerate\commands\launch.py", line 785, in simple_launcher
    raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
subprocess.CalledProcessError: Command '['E:\\Code\\musubi-tuner-scripts\\.venv\\Scripts\\python.exe', './musubi-tuner/qwen_image_train_network.py', '--dataset_config=./toml/qinglong-qwen-image-edit-datasets.toml', '--dit=./ckpts/diffusion_models/qwen_image_edit_2509_bf16.safetensors', '--vae=./ckpts/vae/qwen_image_vae.safetensors', '--seed=1026', '--learning_rate=2e-4', '--output_name=qwen_image_edit_lora_qinglong', '--output_dir=./output_dir', '--logging_dir=./logs', '--text_encoder=./ckpts/text_encoder/qwen_2.5_vl_7b.safetensors', '--edit_plus', '--flash_attn', '--split_attn', '--timestep_sampling=qwen_shift', '--guidance_scale=1', '--max_train_epochs=40', '--gradient_checkpointing', '--network_args', 'loraplus_lr_ratio=4', '--network_dim=32', '--network_alpha=16', '--network_module=networks.lora_qwen_image', '--lr_scheduler=cosine_with_min_lr', '--lr_scheduler_num_cycles=1', '--lr_decay_steps=0.2', '--lr_scheduler_min_lr_ratio=0.1', '--mixed_precision=bf16', '--fp8_base', '--fp8_scaled', '--persistent_data_loader_workers', '--blocks_to_swap=26', '--use_pinned_memory_for_block_swap', '--compile', '--compile_backend=inductor', '--compile_mode=default', '--compile_dynamic', '--compile_cache_size_limit=32', '--img_in_txt_in_offloading', '--optimizer_type=adamw', '--save_every_n_epochs=2', '--sample_at_first', '--sample_every_n_epochs=2', '--sample_prompts=./toml/qinglong_qwen_image_edit.txt']' returned non-zero exit status 1.
```

**Comment by sdbds** [comment]
Upon inspection, this appears to be an issue with the accelerator and the QwenImageTransformer2DModel property checking.
The simplest way to modify it is to add a try/catch block in the sampling method...
```
       try:
            transformer = accelerator.unwrap_model(transformer)
        except AttributeError:
            pass
```
Should we also add a tf32 option?

kohya-ss:
> Upon inspection, this appears to be an issue with the accelerator and the QwenImageTransformer2DModel property checking.
> The simplest way to modify it is to add a try/catch block in the sampling method...

This error seems to occur in newer versions of accelerate. Since we're compiling without using the accelerate function, it seems to be behaving unintentionally. I've made an ad-hoc solution, so it should work now.

> Should we also add a tf32 option?

The speed improvement was not great, but being able to change it as an option would be very useful.

FurkanGozukara:
> --compile_dynamic

thanks i made it work with that info @kohya-ss 

--compile --compile_dynamic : 4.68 second / it

--compile  : 4.63 second / it

it didn't make big difference. what is the benefit of --compile_dynamic ? 

my dataset is all 1328x1328 and bs1

can you please also add args to start timing after 1st step is done? so calculating average speed after very 1st step is done - or perhaps reset statistics after very first step done, whichever easier for you. but very first step is breaking entire speed metric since it starts from start command - includes loading models preparation so many unrelated parts duration

kohya-ss:
> it didn't make big difference. what is the benefit of --compile_dynamic ?

If dynamic is not specified, compilation is required every time the input shape changes. This means that compilation may be required for every step in the first epoch (and subsequent epochs if the batch size is 2 or more). If dynamic=True, compilation is required only for the first step.

> can you please also add args to start timing after 1st step is done? so calculating average speed after very 1st step is done - or perhaps reset statistics after very first step done, whichever easier for you. but very first step is breaking entire speed metric since it starts from start command - includes loading models preparation so many unrelated parts duration

I will work on another PR sooner.

arledesma:
@kohya-ss might want to link out to the torch docs to guide users for the compiler requirements:

https://docs.pytorch.org/tutorials/unstable/inductor_windows.html#install-a-compiler
> Install a Compiler
> C++ compiler is required for TorchInductor optimization, let’s take Microsoft Visual C++ (MSVC) as an example.
> 
> Download and install [MSVC](https://visualstudio.microsoft.com/downloads/).
>
> During Installation, select Workloads and then Desktop & Mobile. Select a checkmark on Desktop Development with C++ and install.
>  


https://docs.pytorch.org/tutorials/intermediate/torch_compile_tutorial.html#introduction-to-torch-compile
> System requirements - A C++ compiler, such as g++ - Python development package (python-devel/python-dev)

Went through similar troubleshooting with @Sarania over in https://github.com/kohya-ss/musubi-tuner/discussions/232#discussioncomment-14204687 and I'm sure that many casual users of this tool will attempt to use this feature and create noisy issues to troubleshoot their environment.

kohya-ss:
> @kohya-ss might want to link out to the torch docs to guide users for the compiler requirements:

Thank you! This is really important information, I'll add it to the documentation.

dxqb:
> > it didn't make big difference. what is the benefit of --compile_dynamic ?
> 
> If dynamic is not specified, compilation is required every time the input shape changes. This means that compilation may be required for every step in the first epoch (and subsequent epochs if the batch size is 2 or more). If dynamic=True, compilation is required only for the first step.
> 

dynamic=None isn't the same as dynamic=False. At dynamic=None it supposedly auto-detects when to go dynamic. Only at dynamic=False you get the described behaviour that it re-compiles each new sequence length.
dynamic=True forces dynamic.

Padding the sequence length can help to avoid recompiles, especially if you remove the padding by split-attention afterwards anyway:
https://github.com/Nerogar/OneTrainer/blob/2e682dd2fbbe4712b5a081fa3b2d3e767b92291a/modules/model/ChromaModel.py#L221

Be prepared for various compatibility issues and bugs when this gets wider use:
https://github.com/Nerogar/OneTrainer/issues/1126
https://github.com/Nerogar/OneTrainer/issues/1120
https://github.com/pytorch/pytorch/issues/165506
https://github.com/pytorch/pytorch/issues/163876
https://github.com/pytorch/pytorch/issues/167891
https://github.com/pytorch/pytorch/issues/167082

happy compiling

FurkanGozukara:
Ok I have tested newest merge

I did reset timing after 1st step to get accurate timings otherwise impossible until you get like 500 steps done

image

5.37 second / it > 4.51 second / it = 16% speed up

Torch 2.8 results

Qwen Image LoRA rank 64 , 1328x1328, batch size 1

`No torch compile`

image

`--compile --compile_dynamic --compile_mode max-autotune-no-cudagraphs`

image

Torch 2.9.1 results

Torch 2.9.1 with CUDA 13 made no improvements

`--compile --compile_dynamic --compile_mode max-autotune-no-cudagraphs`

image

FurkanGozukara:
@kohya-ss i hope you can add args of start timing / statistics after 1st step done so people can see actual speed and ETA from beginning not like after 500 steps. thank you

kohya-ss:
> dynamic=None isn't the same as dynamic=False. At dynamic=None it supposedly auto-detects when to go dynamic. Only at dynamic=False you get the described behaviour that it re-compiles each new sequence length. dynamic=True forces dynamic.

Thank you for pointing that out! I didn't understand this part. I'd like to update the options.

> Padding the sequence length can help to avoid recompiles, especially if you remove the padding by split-attention afterwards anyway: https://github.com/Nerogar/OneTrainer/blob/2e682dd2fbbe4712b5a081fa3b2d3e767b92291a/modules/model/ChromaModel.py#L221

Thank you for this too. We are currently padding, but we were padding to max_seq_len. Making it 16 units is a good idea.

Also, the information you provided about the bug was extremely helpful. These are annoying problems...

dxqb:
> > dynamic=None isn't the same as dynamic=False. At dynamic=None it supposedly auto-detects when to go dynamic. Only at dynamic=False you get the described behaviour that it re-compiles each new sequence length. dynamic=True forces dynamic.
> 
> Thank you for pointing that out! I didn't understand this part. I'd like to update the options.
> 
> > Padding the sequence length can help to avoid recompiles, especially if you remove the padding by split-attention afterwards anyway: https://github.com/Nerogar/OneTrainer/blob/2e682dd2fbbe4712b5a081fa3b2d3e767b92291a/modules/model/ChromaModel.py#L221
> 
> Thank you for this too. We are currently padding, but we were padding to max_seq_len. Making it 16 units is a good idea.

Padding to max length can be better if you remove that pads by split-attention anyway. If an attention mask is used, padding to max length is very slow that's why I chose 16.
Although, padding to max length could also be slower in linear layers, where you don't split, right? Can be significant in low-res training. At 512px image sequence length is only 1024. Plus max text sequence length 512?

kohya-ss:
> Padding to max length can be better if you remove that pads by split-attention anyway. If an attention mask is used, padding to max length is very slow that's why I chose 16.
> Although, padding to max length could also be slower in linear layers, where you don't split, right? Can be significant in low-res training. At 512px image sequence length is only 1024. Plus max text sequence length 512?

Sorry, I omitted a bit too much. I'm padding to `max(txt_seq_lens)` where `txt_seq_lens` is the lengthes of the text sequence in each element of the batch :)

Sarania:
Very nice, it's excellent to see this supported in mainline!  That said when merging it to blissful-tuner, I noticed it seemed significantly less effective in some cases. As an example - for Hunyuan Video, generate 960x720x121f, fp8_scaled, fp8_fast, sageattn, 20 blocks swapped, my implementation will do it with about 15GB/16GB used on my GPU. Same settings, replace my compile with this one(same compile settings):

OutOfMemoryError: CUDA out of memory. Tried to allocate 1.92 GiB. GPU 0 has a total capacity of 15.81 GiB of which 1.37 GiB is free

so it seems to be using like 1.5 more gigs. The only real difference I can note is:

disable_linear=blocks_to_swap > 0

what's the reason for doing this? I think it must be why as it's the only notable difference in the implementations.

Edit: Yeah if I comment the disable_linear out, your version just slightly outperforms mine at 14.97GB use! So I would like to understand it's purpose to know what I wanna do about this in blissful ☺️ If you could explain that would be awesome!

kohya-ss:
> disable_linear=blocks_to_swap > 0
> 
> what's the reason for doing this? I think it must be why as it's the only notable difference in the implementations.

This is because when block swap is applied and compile is enabled for Linear, the backward hook does not work properly when tracing for compilation. Therefore, it seems that setting this to False will not be a problem during inference. Thanks for pointing this out!

However, when I tried disabling it during training, it worked fine for some reason for now. I will check further. It may be disabled even in training.

Sarania:
I've learned that compile can be really tricky especially once you start throwing in different versions of torch, multiplatform stuff and different systems. And it's hard for me to debug that because on my system it tends to #JustWork a lot of the time. It might be worth making just an optional switch for it this way if any users do encounter the problem when training, it would be available. But yeah it shouldn't be a problem during inference, I've had compile enabled(well enable-able at least) for all models in blissful on the inference side for a while now and not noticed any issues, but training is definitely trickier. I created an internal optimized compile specifically for Wan training that works heckin' beautifully for me on Linux but gives others lots of issues especially on Windows. I like this implementation you've done overall it's cleaner than mine so ultimately I'm going to replace what I've done with it for the most part, that's why I wanted to understand why that bit was happening and I figured it was something like that thanks for explaining!

Sarania:
So for blissful I've currently decided I will default to compiling the linears for inference and not compiling them for training(unless no block swap), but I add switches to reverse this behavior in either case. On my system at least, even with block swap enabled compiling them for training works fine, so if it proves to not be much of an issue it might be worth looking at again in the future because it does make a noticeable difference in VRAM usage!

PR #712, opened 2025-11-05
Add LongCat-Video LoRA training

AkaneTendo25:
Needs more testing, but already functional. 
I made a training test run (T2V) on a 512x512 image dataset (20 items), RTX 3090, 64 gb RAM (Python 3.10, PyTorch 2.6.0, FlashAttn 2.7.4).
Used bf16 checkpoint, dynamic quantization with `--fp8_base`, `--fp8_scaled` flags.
Samples that I got during training confirm model learned style properly (not showing here, cause slightly NSFW).
The speed was approximately 4s/it.

<details><summary>Loss graph</summary>
<p>
image
</p>
</details> 

Here were the exact commands and configs used for training (all launched within activated venv):

<details><summary>Cache latents</summary>
<p>

`python longcat_cache_latents.py --dataset_config _longcat_lora_dataset.toml --vae E:/ComfyUI_windows_portable/ComfyUI/models/vae/wan/wan_2.1_vae.safetensors`

</p>
</details> 

<details><summary>Cache encoder outputs</summary>
<p>

`python longcat_cache_text_encoder_outputs.py --dataset_config _longcat_lora_dataset.toml --text_encoder H:/models/wan21/Wan2.1-T2V-14B/models_t5_umt5-xxl-enc-bf16.pth --batch_size 16 `

</p>
</details> 

<details><summary>Dataset config</summary>
<p>

```
[general]
enable_bucket = true
bucket_no_upscale = false

[[datasets]]
image_directory = "H:/datasets/moebius_test" 
cache_directory = "H:/datasets/moebius_test/cache" 
resolution = [512, 512] 
batch_size = 1 
num_repeats = 1 
caption_extension = ".txt"
```

</p>
</details> 

<details><summary>Sampling prompts</summary>
<p>

```
# prompt 1
Moebius style. Woman with blonde hair is walking on the beach, camera slowly zooms out revealing her naked body.  --w 384 --h 384 --f 45 --d 4 --s 20
```

</p>
</details> 

<details><summary>Training</summary>
<p>

```
accelerate launch --num_cpu_threads_per_process 1 --mixed_precision bf16 longcat_train_network.py ^
    --dit F:/LongCat_full_bf16.safetensors ^
    --vae E:/ComfyUI_windows_portable/ComfyUI/models/vae/wan/wan_2.1_vae.safetensors ^
    --text_encoder H:/models/wan21/Wan2.1-T2V-14B/models_t5_umt5-xxl-enc-bf16.pth ^
    --flash_attn ^
[...]

PR #774, opened 2025-12-11
Add Kandinsky 5 Pro LoRA training

AkaneTendo25:
@kohya-ss 

Sorry for submitting yet another PR before finishing my previous one ([for LongCat-Video](https://github.com/kohya-ss/musubi-tuner/pull/712)), I will take care of the latter as well soon. But I wanted to be the first to contribute K5 training pipeline to your wonderful project ❤️
I will not abandon this PR and plan to continue pushing it toward production readiness.

The Kandinsky-5 Pro model is promising, and thanks to the ComfyUI team ([especially Kijai](https://github.com/comfyanonymous/ComfyUI/pull/10988)) it should soon be available to a wide audience in the main ComfyUI release (and probably should run fast enough with distillation LoRA). I hope users will try it and will love to create some great LoRAs for it.

This is a draft version, and I've only tested T2V LoRA, but both the training and inference pipelines are working. With musubi-tuner optimizations (fp8 and/or block swapping), video training fits well within consumer setup, with tolerable (expected for a 19B model) performance. With 10 blocks swapped I get 5.5s/it with mixed (512p images + 256p/17f videos) dataset on Windows 11, 24 Gb VRAM (RTX 3090), DDR4 128 Gb RAM (64 Gb probably will be enough, but cannot confirm). Pretrain-5s model fine-tunes reliably and demonstrates strong style capture. 

Training and inference examples:

<details><summary>Details</summary>
<p>

VAE latents cache:
```
python kandinsky5_cache_latents.py ^
  --dataset_config G:/samples/kandinsky-tuner/_kandinsky_lora_dataset.toml ^
  --vae F:/k5_modules/hunyuan/vae/diffusion_pytorch_model.safetensors ^
  --device cuda ^
  --vae_dtype fp16
```

Text embeddings cache:
```
python kandinsky5_cache_text_encoder_outputs.py ^
  --dataset_config G:/samples/kandinsky-tuner/_kandinsky_lora_dataset.toml ^
  --text_encoder_qwen F:/k5_modules/text_encoder ^
  --text_encoder_clip F:/k5_modules/text_encoder2 ^
  --device cuda ^
  --quantized_qwen ^
  --batch_size 4
```

Training (T2V):
```
accelerate launch --num_cpu_threads_per_process 1 --mixed_precision bf16 kandinsky5_train_network.py ^
  --mixed_precision bf16 ^
  --dataset_config G:/samples/kandinsky-tuner/_kandinsky_lora_dataset.toml ^
  --task k5-pro-t2v-5s-sd ^
  --dit F:/kandinsky5pro_t2v_pretrain_5s.safetensors ^
  --text_encoder_qwen F:/k5_modules/text_encoder ^
  --text_encoder_clip F:/k5_modules/text_encoder2 ^
  --vae F:/k5_modules/hunyuan/vae/diffusion_pytorch_model.safetensors ^
  --fp8_base ^
  --blocks_to_swap 10 ^
  --flash_attn ^
  --gradient_checkpointing ^
  --max_data_loader_n_workers 1 ^
  --persistent_data_loader_workers ^
  --learning_rate 1e-4 ^
  --optimizer_type AdamW8Bit ^
  --optimizer_args "weight_decay=0.001" "betas=(0.9,0.95)" ^
  --max_grad_norm 1.0 ^
  --lr_scheduler constant_with_warmup ^
  --lr_warmup_steps 100 ^
  --network_module networks.lora_kandinsky ^
  --network_dim 16 ^
  --network_alpha 16 ^
  --timestep_sampling shift ^
  --discrete_flow_shift 5.0 ^
  --output_dir G:/samples/kandinsky-tuner/output ^
  --output_name kandinsky_lora_moebius ^
  --log_config ^
  --log_with tensorboard ^
  --logging_dir G:/samples/kandinsky-tuner/logs ^
  --save_state ^
  --save_every_n_epochs 5 ^
  --max_train_epochs 50 ^
  --sample_prompts G:/samples/kandinsky-tuner/_kandinsky_lora_sampling.txt ^
  --scheduler_scale 10.0 ^
  --offload_dit_during_sampling ^
  --sample_every_n_epochs 5
```

Generate sample video (T2V):
```
python kandinsky5_generate_video.py ^
  --task k5-pro-t2v-5s-sd ^
  --dit F:/kandinsky5pro_t2v_pretrain_5s.safetensors ^
  --vae F:/k5_modules/hunyuan/vae/diffusion_pytorch_model.safetensors ^
  --text_encoder_qwen F:/k5_modules/text_encoder ^
  --text_encoder_clip F:/k5_modules/text_encoder2 ^
  --blocks_to_swap 10 ^
  --offload_dit_during_sampling ^
  --fp8_base ^
  --dtype bfloat16 ^
  --prompt "Moebius style, woman with blonde hair walking on the beach, camera slowly zooms out." ^
  --negative_prompt "low quality, artifacts" ^
  --frames 17 ^
  --steps 10 ^
  --guidance 5 ^
  --scheduler_scale 10 ^
  --seed 42 ^
  --width 384 ^
  --height 384 ^
  --lora_weight G:/samples/kandinsky-tuner/output/kandinsky_lora_moebius-000010.safetensors ^
  --lora_multiplier 1.0 ^
  --output output\k5_sample_lora.mp4
```

</p>
</details> 

Roadmap (remaining tasks):

- [x] Implement I2V training
- [x] Fix incorrect AI-generated documentation
- [x] Add [NABLA](https://arxiv.org/abs/2507.13546v1) [sparse attention](https://github.com/gen-ai-team/Wan2.1-NABLA) for fine-tuning CrossDiT (see also [K5 tech report](https://arxiv.org/abs/2511.14993), section 5.4)
- [x] (optional) Verify pipeline compatibility with K5-Lite video models
- [x] <del>(optional) Add support for T2I and I2I (editing) models</del> currently not planned
- [x] Remove unused code
- [x] Resolve issues found during review

Training with NABLA has been implemented, but I haven't been able to practically validate it. NABLA's benefits are expected mainly at higher video resolutions and longer durations, which I can't currently run on my hardware.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,15 +109,6 @@ extend-exclude = [
   "src/musubi_tuner/frame_pack/k_diffusion_hunyuan.py",
   "src/musubi_tuner/hunyuan_model/autoencoder_kl_causal_3d.py",
   "src/musubi_tuner/hunyuan_model/pipeline_hunyuan_video.py",
-  "src/musubi_tuner/kandinsky/models/__init__.py",
-  "src/musubi_tuner/kandinsky/models/attention.py",
-  "src/musubi_tuner/kandinsky/models/dit.py",
-  "src/musubi_tuner/kandinsky/models/nn.py",
-  "src/musubi_tuner/kandinsky/models/text_embedders.py",
-  "src/musubi_tuner/kandinsky/models/utils.py",
-  "src/musubi_tuner/kandinsky/models/vae.py",
-  "src/musubi_tuner/kandinsky/__init__.py",
-  "src/musubi_tuner/kandinsky/generation_utils.py",
   "src/musubi_tuner/modules/scheduling_flow_match_discrete.py",
   "src/musubi_tuner/modules/unet_causal_3d_blocks.py",
   "src/musubi_tuner/qwen_image/qwen_image_autoencoder_kl.py",
@@ -163,15 +154,6 @@ docstring-code-line-length = "dynamic"
 "src/musubi_tuner/frame_pack/k_diffusion_hunyuan.py" = ["ALL"]
 "src/musubi_tuner/hunyuan_model/autoencoder_kl_causal_3d.py" = ["ALL"]
 "src/musubi_tuner/hunyuan_model/pipeline_hunyuan_video.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/models/__init__.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/models/attention.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/models/dit.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/models/nn.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/models/text_embedders.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/models/utils.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/models/vae.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/__init__.py" = ["ALL"]
-"src/musubi_tuner/kandinsky/generation_utils.py" = ["ALL"]
 "src/musubi_tuner/modules/scheduling_flow_match_discrete.py" = ["ALL"]
 "src/musubi_tuner/modules/unet_causal_3d_blocks.py" = ["ALL"]
 "src/musubi_tuner/qwen_image/qwen_image_autoencoder_kl.py" = ["ALL"]


kabachuha:
I wonder what can even be trained on `256p/17f` 😬

kohya-ss:
Thank you for this! The Kandinsky-5 Pro model looks promising. I will review it as soon as it is ready.

AkaneTendo25:
> I wonder what can even be trained on `256p/17f` 😬

@kabachuha Here this was just a smoke test 🙂. In general, however, this resolution is fine for training motion LoRAs.
For example, I trained Wan (for style, but it included some specific 2D animation patterns) using a combination of 256x144x81 and 512x288x17 clips, along with 976x544 images, on an RTX 3090 (without block swapping), and the results were decent.

maybleMyers:
This really is a great model, much better fidelity and smoother than wan2.2. 
Kohya I have been working on an inference script, it might give you somewhere to start from with more features than the test script https://github.com/maybleMyers/kandinsky5/blob/main/kandi_generate_video.py
Currently v2v needs some work but most other options work okay.

kabachuha:
@AkaneTendo25 Thank you, works magnificently! 🫡

Launched it at 256x256x72 frames with 20 block swap on 5090, and fits within 24 gbs with ~6 s/it! 

https://github.com/user-attachments/assets/1f075bb5-e0b6-4f17-b5fa-8967bab46e09

Two notes:

1. In your documentation, you need to change the Qwen2.5 VL 3B to Qwen2.5 VL 7B.
2. Your encoder points to clip_l.safetensors, but the actual encoding script demands a transformers model, and you need to point to openai/clip-vit-large-patch14-336 (or the cached one)

Also, to load the checkpoints in ComfyUI a small correction needs to be made for the conver_lora script:

```python
if "visual.transformer.blocks" in module_name or "text.transformer.blocks" in module_name: # Kandinsky 5
    module_name = module_name.replace("visual.transformer.blocks", "visual_transformer_blocks")
    module_name = module_name.replace("self.attention", "self_attention")
    module_name = module_name.replace("cross.attention", "cross_attention")
    module_name = module_name.replace("feed.forward", "feed_forward")
    module_name = module_name.replace("to.", "to_")
    module_name = module_name.replace("in.", "in_")
    module_name = module_name.replace("out.", "out_")
    module_name = module_name.replace("text.transformer.blocks", "text_transformer_blocks")
```

I tuned it on the image to video checkpoint directly and it works out of the box if you just point it to its path. (Thankfully, no additional cross-attention mechanisms are required like in Wan2.1 or HunyuanVideo versions)

AkaneTendo25:
@kabachuha That's amazing to hear, thank you very much for testing this! 

Regarding CLIP, you're right, thanks for pointing this out. That value came from an older version of my training script, but it didn't cause any errors because the training uses pre-cached embeddings generated with a manual caching script.

AkaneTendo25:
@kabachuha I forgot to mention: right now I2V training technically runs but the pipeline ignores the source image: `visual_cond` is filled with zeros, so the output LoRA should behave like T2V LoRA. In the next update (that I will hopefully push tomorrow) it will require `latents_image` (cached first-frame latents) and actually use them, so the LoRA will condition on images.

kabachuha:
Yes, I know the model's structure/latent inputs 😀 (after almost implementing it in diff-pipe)

I'm just very impatient in this regard.

Afaik, the visual cond simply replaces the first one in the noisy batch, and it's not a big deal in the structure (unlike cross-attn/image embedding concat), for learning motions and concepts

kabachuha:
@AkaneTendo25 Does FLF2V work in your experience?

How would one implement it in inference like in ComfyUI?

AkaneTendo25:
@kabachuha Tbh I haven't yet tested training a LoRA specifically for this method, but based on the architecture there's a good chance it can work. As for inference in ComfyUI, I haven't thought about it yet, but if there's no dedicated node for this (like WanFirstLastFrameToVideo), I'll try to make my own.

AkaneTendo25:
@kohya-ss I believe the code is ready for review.
The Japanese version of the Kandinsky documentation is AI-generated, so there may be some inaccuracies.

Please feel free to take a look whenever it's convenient for you.

kabachuha:
@AkaneTendo25 I tried this one (adding also the code that replaced the last latent of the videos with the end frame), but it didn't work 😭

```python
class Kandinsky5FirstLastFrameToVideo(io.ComfyNode):
    @classmethod
    def define_schema(cls):
        return io.Schema(
            node_id="Kandinsky5FirstLastFrameToVideo",
            category="conditioning/video_models",
            inputs=[
                io.Conditioning.Input("positive"),
                io.Conditioning.Input("negative"),
                io.Vae.Input("vae"),
                io.Int.Input("width", default=768, min=16, max=nodes.MAX_RESOLUTION, step=16),
                io.Int.Input("height", default=512, min=16, max=nodes.MAX_RESOLUTION, step=16),
                io.Int.Input("length", default=121, min=1, max=nodes.MAX_RESOLUTION, step=4),
                io.Int.Input("batch_size", default=1, min=1, max=4096),
                io.Image.Input("start_image", optional=True),
                io.Image.Input("end_image", optional=True),
            ],
            outputs=[
                io.Conditioning.Output(display_name="positive"),
                io.Conditioning.Output(display_name="negative"),
                io.Latent.Output(display_name="latent", tooltip="Empty video latent"),
                io.Latent.Output(display_name="cond_latent_start", tooltip="Clean encoded start images, used to replace the noisy start of the model output latents"),
                io.Latent.Output(display_name="cond_latent_end", tooltip="Clean encoded end images, used to replace the noisy start of the model output latents"),
            ],
        )

    @classmethod
    def execute(cls, positive, negative, vae, width, height, length, batch_size, start_image=None, end_image=None) -> io.NodeOutput:
        latent = torch.zeros([batch_size, 16, ((length - 1) // 4) + 1, height // 8, width // 8], device=comfy.model_management.intermediate_device())
        cond_latent_start = {}
        cond_latent_end = {}
        conditioning_values = {}

        conditioning_values["concat_mask"] = torch.ones((1, 1, latent.shape[2], latent.shape[-2], latent.shape[-1]), device=latent.device, dtype=latent.dtype)

        if start_image is not None:
            start_image = comfy.utils.common_upscale(start_image[:length].movedim(-1, 1), width, height, "bilinear", "center").movedim(1, -1)
            encoded = vae.encode(start_image[:, :, :, :3])
            cond_latent_start["samples"] = encoded
            conditioning_values["concat_mask"][:, :, :((start_image.shape[0] - 1) // 4) + 1] = 0.0
            conditioning_values["time_dim_replace"] = encoded

        if end_image is not None:
            end_image = comfy.utils.common_upscale(end_image[:length].movedim(-1, 1), width, height, "bilinear", "center").movedim(1, -1)
            encoded = vae.encode(end_image[:, :, :, :3])
            cond_latent_end["samples"] = encoded
            conditioning_values["concat_mask"][:, :, -(((end_image.shape[0] - 1) // 4) + 1):] = 0.0
            conditioning_values["time_dim_replace_end"] = encoded

        positive = node_helpers.conditioning_set_values(positive, conditioning_values)
        negative = node_helpers.conditioning_set_values(negative, conditioning_values)

        out_latent = {}
        out_latent["samples"] = latent
        return io.NodeOutput(positive, negative, out_latent, cond_latent_start, cond_latent_end)
```

AkaneTendo25:
@kabachuha When you say "didn't work", do you mean (1) the node fails to run, or (2) it runs but the model doesn't actually follow the end-frame condition? If it's (2), isn't that expected, since in the I2V the model is explicitly trained to condition on the first frame only, and the report doesn't describe any native last-frame conditioning. So a pure inference-time hack (injecting the last latent) may be just ignored.

My initial hope is that FLF2V could be learnable as a downstream task via LoRA, since the model already supports I2V-style conditioning, and if a model can be adapted to interpret temporal masks more generally, FLF behavior may be inducible as a latent capability rather than requiring a fully separate supervised task.

But this remains a hypothesis at this point, I haven't yet trained a LoRA specifically for FLF2V to verify this.

kabachuha:
> So a pure inference-time hack (injecting the last latent) may be just ignored.

Yes, node absolutely runs (with time_dim_replace_end hacked into the latents inside the model's code), and it simply results in noisy-result in-between. See how the video starts and ends with slowly noised latents and then noise at the center.

https://github.com/user-attachments/assets/04a316ff-b382-4b5e-9814-2d259b31525a

> My initial hope is that FLF2V could be learnable as a downstream task via LoRA, since the model already supports I2V-style conditioning

Maybe a larger training is needed. I'd bet more on LyCORIS instead of LoRA, because it's thought to be superior for larger capabilities fine-tuning.

---

Btw, I spotted that they already have a FLF version in SberbankOnline's app when they presented the financial report of the year: looped videos (with preset images, not something like Moebius) and obvious transitions between two given frames. Let's hope they'll release this version eventually 😀

kohya-ss:
Thank you! I will review this tomorrow!

kohya-ss:
All of the Copilot reviews are minor details so I think you can ignore them.

docs/kandinsky5.md
```diff
@@ -0,0 +1,471 @@
...
+**Pro models (19B parameters):**
+
+| Task | Description | Resolution |
+|------|-------------|------------|
+| `k5-pro-t2v-5s-sd` | Pro T2V 5s SD | 512 |
+| `k5-pro-t2v-5s-hd` | Pro T2V 5s HD | 1024 |
+| `k5-pro-t2v-10s-sd` | Pro T2V 10s SD | 512 |
+| `k5-pro-t2v-10s-hd` | Pro T2V 10s HD | 1024 |
+| `k5-pro-i2v-5s-sd` | Pro I2V 5s SD | 512 |
+| `k5-pro-i2v-5s-hd` | Pro I2V 5s HD | 1024 |
```
kohya-ss: I would appreciate it if you could tell us which weights to use for which tasks.

docs/kandinsky5.md
```diff
@@ -0,0 +1,471 @@
...
+
+`--fp8_base` runs DiT in fp8 mode. This can significantly reduce memory consumption but may impact output quality.
+
+If you're running low on VRAM, use `--blocks_to_swap` to offload some blocks to CPU.
+
+`--gradient_checkpointing_cpu_offload` can be used to offload activations to CPU when using gradient checkpointing. This must be used together with `--gradient_checkpointing`.
+
+### Attention / アテンション
+
+Use `--sdpa` for PyTorch's scaled dot product attention. Use `--flash_attn` for FlashAttention. Use `--xformers` for xformers.
```
kohya-ss: The code seems to always use SDPA (ignoring `--flash_attn` and `--xformers`).

kohya-ss:
In this repository, the options for the weight files use a format that directly specifies the weight safetensors rather than the HuggingFace ID. This is also to make the script compatible with ComfyUI weights. Therefore, I will update this after merging.

Also, since a setting of 1 is recommended for `--network_alpha` (this requires changing the learning rate, but should not affect the training itself), I may update the documentation accordingly.

Also, I would appreciate it if you could tell me when to use Nabla and when not to use it.

In my understanding, if we don't use Nabla, the script will use SDPA with attention mask, so training will likely take a little longer than other models.

AkaneTendo25:
@kohya-ss Thank you for the review!

I will fix the issue where non-SDPA attention engines are not used, I somehow missed that.

Regarding NABLA, here is what I found out. Per the Kandinsky‑5 [tech report](https://arxiv.org/abs/2511.14993), NABLA is meant to speed up training by dynamically constructing attention masks, and its benefits mainly show up at higher resolutions and/or longer durations. I haven't been able to test it extensively because my local hardware can't handle those settings.

I have no evidence that Kandinsky‑5 developers tested NABLA for LoRA training, so its effect in low‑rank fine‑tuning is still unknown. I added it primarily to enable experimentation, and the code was mostly taken from original [Kandinsky-5 repo](https://github.com/kandinskylab/kandinsky-5) and their "official" [trainer](https://github.com/kandinskylab/kandinsky-5-lora-train). If you prefer, I can remove NABLA support for now and re‑introduce it later in a separate PR once there would be a stronger validation of its effect.

What I wanted to mention explictly is that NABLA sets some constraints for input resolutions.  I'm including the breakdown below to justify the `--nabla_resize` flag in the caching CLI and to make it easy to trace (and correct) my logic if needed.

The NABLA mask builder method (`nablaT_v2`) chunks the token sequence into blocks of 64 (`S // 64`) and uses `BlockMask(..., BLOCK_SIZE=64`), so the total visual token count `S` must be divisible by 64. This requirement comes (I think) from the Kandinsky-5 paper which says (section 5.4):

> NABLA employs token reordering through fractal flattening with spatial patches of size P x P (P = 8), grouping all tokens within each patch into contiguous sequences of P^2 = N = 64 tokens

To guarantee that, the code groups tokens into 8 x 8 spatial tiles in `fractal_flatten()` (hard-coded `pixel_size = 8`). If `H′` and `W′` (the spatial token grid after patching) are multiples of 8, then `H′ x W′` is a multiple of 64 and `S = T x H′ x W′` is divisible by 64 for any `T` (number of time tokens = number of frames after temporal patching).

In Kandinsky‑5:
- VAE downsamples spatially by 8, so `latent_h = input_h / 8`
- DiT patch size is `(1, 2, 2)` (2x spatial patching) in `src/musubi_tuner/kandinsky5/configs.py`, so `H′ = latent_h / 2`

So `H′ % 8 == 0`, and `latent_h % 16 == 0`, and `input_h % 128 == 0` (same for width).

That's why `--nabla_resize` in `kandinsky5_cache_latents.py` is introduced - it resizes inputs to multiples of 128, which guarantees latent `H/W` divisible by 16, token grid divisible by 8, and therefore `S` divisible by 64.

src/musubi_tuner/kandinsky5_generate_video.py
```diff
@@ -0,0 +1,300 @@
...
+    if neg_attention_mask is not None:
+        neg_attention_mask = neg_attention_mask.to("cpu")
+    if attention_mask is not None:
+        mask = attention_mask[0] if attention_mask.dim() > 1 else attention_mask
+        mask = mask.bool().flatten()
+        if mask.shape[0] != text_embeds.shape[0]:
+            raise ValueError(
+                f"attention_mask length {mask.shape[0]} does not match text_embeds length {text_embeds.shape[0]}; "
+                "please re-cache Kandinsky5 text embeddings"
+            )
```
kohya-ss: It seems that I am getting the following error: `ValueError: attention_mask length 256 does not match text_embeds length 11; please re-cache Kandinsky5 text embeddings`

Truncating the mask like this worked: `mask = mask[:text_embeds.shape[0]]`

AkaneTendo25:
@kohya-ss Sorry for this, I forgot to check generate video script after I changed attention mask processing during training (and this logic is shared by generate script).
Should be fixed now in https://github.com/kohya-ss/musubi-tuner/pull/774/commits/2897e17133a04471b9bdfa49e2f41d1d2f3a5dee

kohya-ss:
Thank you for update! the generation script worked. I didn't have time to train LoRA properly, but I have confirmed that the content changes when applying trained LoRA.

Could you please format the code with ruff?

AkaneTendo25:
Fixed it.

I also noticed a mistake in pyproject.toml: I added some scripts under src/musubi_tuner/kandinsky/ to the Ruff exception list, but the actual path is src/musubi_tuner/kandinsky**5**/.

Since all Kandinsky scripts now appear to be fully Ruff-compliant, would you prefer that I remove the Kandinsky-related exclusions from pyproject.toml altogether?

kohya-ss:
Thank you! If the scripts are already ruff-compliant, you can remove the exception from your pytproject.toml.

(I did that for HunyuanVideo 1.5.)

kohya-ss:
Thank you again for this great PR! Sorry it took so long.

AkaneTendo25:
Thank you very much for your time! 
If any issues related to the Kandinsky pipeline come up, I'll be glad to help with fixing it.

PR #781, opened 2025-12-13
Feat enable full bf16 for z image train

kohya-ss:
fp16/bf16 training is accidentally enabled without `full_bf16` option. This PR makes the default precision fp32, and enables `full_bf16` option to work.

Also add the documentation for Z-Image fine tuning.


Krishicall:
Hello, what are the usage instructions for full_bf16? Throwing in the --full_bf16 flag shows an argument error for me. Updated my main branch this morning. Thanks!

PR #674, opened 2025-10-23
Fixing multiple GPU Qwen Image Fine tuning training

FurkanGozukara:
I am not 100% sure if this is accurate way but after this i was able to train

Haven't tested result of training yet

This fixes issue mentioned here : https://github.com/kohya-ss/musubi-tuner/issues/672

PR #793, opened 2025-12-19
Fix DDP bugs

sdbds:
When using DDP with multiple GPUs, the model is wrapped without obtaining the underlying properties.

### DIFF
--- a/src/musubi_tuner/zimage_train_network.py
+++ b/src/musubi_tuner/zimage_train_network.py
@@ -179,11 +179,11 @@ class ZImageNetworkTrainer(NetworkTrainer):
             latent_model_input = latent_model_input.unsqueeze(2)  # Add frame dimension [B, C, F, H, W]
 
             with accelerator.autocast(), torch.no_grad():
-                model_out = model(latent_model_input, timestep, embed, mask)
+                model_out = transformer(latent_model_input, timestep, embed, mask)
 
             if do_cfg:
                 with accelerator.autocast(), torch.no_grad():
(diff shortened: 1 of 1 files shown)


src/musubi_tuner/zimage_train_network.py
```diff
@@ -267,7 +267,7 @@ def call_dit(
         timesteps: torch.Tensor,
         network_dtype: torch.dtype,
     ):
-        model: zimage_model.ZImageTransformer2DModel = transformer
+        model: zimage_model.ZImageTransformer2DModel = accelerator.unwrap_model(transformer)
```
kohya-ss, line 270: If we unwrap the model here, the accelerate wrapper will not work when calling the model during training, which may have side effects.

kohya-ss:
Thank you for this! It's strange because DDP training works in my environment. 

However, I don't think there will be any particular negative impact with this update, so I plan to merge it. I would appreciate it if you could check just one place.

src/musubi_tuner/zimage_train_network.py
sdbds: > If we unwrap the model here, the accelerate wrapper will not work when calling the model during training, which may have side effects.

It really is, I just committed the change, and now all the call model weights are changed to the original warp version.

sdbds:
> Thank you for this! It's strange because DDP training works in my environment.
> 
> However, I don't think there will be any particular negative impact with this update, so I plan to merge it. I would appreciate it if you could check just one place.

check #674 same problem with DDP unwarp...so i think now it works.

kohya-ss:
Thank you for update! I will check #674.

PR #795, opened 2025-12-20
Prevent NaN during Z-Image-Turbo fp16 training

yslai:
Port of the NaN prevention method for ComfyUI, consisting of the clamping in the ComfyUI Lumina 2 code, and the scaling method in the pull request by @vanDuven (https://github.com/comfyanonymous/ComfyUI/pull/11187). Currently specifying `--mixed_precision fp16` will result the average loss to become NaN (as opposed to `--mixed_precision bf16`, for example). After the change, fp16 trains stably and also more efficiently on pre-bf16 hardware.

Note that the remaining issues mentioned with ComfyUI is not applicable here, as this is modifies code for Z-Image-Turbo only.

### DIFF
--- a/src/musubi_tuner/zimage/zimage_model.py
+++ b/src/musubi_tuner/zimage/zimage_model.py
@@ -275,7 +275,9 @@ class ZImageTransformerBlock(nn.Module):
             del scale_msa
             x = x + gate_msa * self.attention_norm2(clamp_fp16(attn_out))
             del gate_msa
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you! This works fine.

Could you please format this using ruff? If that's difficult, please let me know and I'll format after merging.

kohya-ss:
Thank you for update!

PR #800, opened 2025-12-22
Step and epoch recovery when resuming

christopher5106:


kohya-ss:
Thank you for this!

The reason why Musubi Tuner (and sd-scripts) does not support restoring steps when resuming is because [accelerate](https://github.com/huggingface/accelerate) did not support saving and restoring steps. In a recent version, accelerate supported saving and restoring steps, so it might be simpler to use that. I will look into how to restore steps with accelerate. If that's difficult, I'll consider merging this pull request.

christopher5106:
Do you have any update about restoring steps ? Thanks

kohya-ss:
Thank you for ping. I'm always busy so I haven't had much time to work on this. I recently confirmed that `global_step` can be restored with accelerate. I'll work on this once FLUX.2 support is complete.

christopher5106:
Hi,
i tried again and global step still not resumed correctly.

christopher5106:
Hi,

I just found where is step recovery in the Accelerate code:

https://github.com/huggingface/accelerate/blob/8067abae81abbd176af65dc9694f9f99dacf3985/src/accelerate/checkpointing.py#L293

https://github.com/huggingface/accelerate/blob/8067abae81abbd176af65dc9694f9f99dacf3985/src/accelerate/accelerator.py#L3849

It's among random state pickle file.

PR #811, opened 2025-12-25
fix(qwen_image_train_network): argument order

jimlee2048:
sync changes  in `qwen_image_train_network` since https://github.com/kohya-ss/musubi-tuner/pull/808/commits/cc03b043328240efbbc8fe8875517b582afe8a7a, otherwise lora training won't work properly.


kohya-ss:
Thank you for this!

PR #812, opened 2025-12-26
Add CUDA 13.0 support

djtuBIG-MaliceX:
# WHAT IS THIS
Using `uv` on Windows 11 with a 16GB RTX5070 GPU installed, this required CUDA 13.0 PyTorch build to work (via `uv pip install torch torchvision --index-url https://download.pytorch.org/whl/cu130 --force-reinstall`).

This simply adds the minimum necessary changes to ensure the trainer via the GUI actually uses the GPU and not the CPU, when executing via `uv run --extra gui .\src\musubi_tuner\gui\gui.py`

Evidence of auto-installed versions via `uv`:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,10 @@ cu128 = [
     "torch>=2.7.1",
     "torchvision>=0.22.1",
 ]
+cu130 = [
+    "torch>=2.9.1",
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for this!

Issue #790, opened 2025-12-16
macOS: bf16 mixed precision requires PyTorch


Gabled-Waters:
I'm one of those annoying Mac users, and I try to use Musubi in conjunction with a ComfyUI workflow (https://github.com/shootthesound/comfyUI-Realtime-Lora) but I don't thing the workflow is to blame. Anyway, I get through creating latents for the images and text alright, but then:

[Musubi Z-Image] Text encoder outputs cached.
[Musubi Z-Image] Starting training: MyLora_20251216_175543
[Musubi Z-Image] Images: 4, Steps: 400, LR: 0.0002, Rank: 16

bf16 mixed precision requires PyTorch >= 1.10 and a supported device.

I tried to opt out on bf16 with accelerate config, and torch is 2.9.1 ... so I guess it is the device part? Or am I missing something?

djtuBIG-MaliceX:
Need to select "no" for Mixed precision under macOS.
At the moment though, there's some code changes required to allow selection of "mps" backend as it's currently hardcoded to select CUDA or CPU only...  And even then, `accelerate` and PyTorch uses `bitsandbytes` operations which are currently not supported on neither MPS nor CPU. (eg: `bitsandbytes::optimizer_update_8bit_blockwise`)

kohya-ss:
#795 is merged, so you may be able to train Z-Image with fp16.

Also, this pull request #801 should remove the hard-coding of the device. I hope you'll give it a try.

djtuBIG-MaliceX:
> [#795](https://github.com/kohya-ss/musubi-tuner/pull/795) is merged, so you may be able to train Z-Image with fp16.
> 
> Also, this pull request [#801](https://github.com/kohya-ss/musubi-tuner/pull/801) should remove the hard-coding of the device. I hope you'll give it a try.

I checked out that branch but I still had to do the following diff since earlier pre-training steps were still selecting "cpu" instead of "mps".  Here's a diff that may help.

```diff
diff --git a/src/musubi_tuner/cache_latents.py b/src/musubi_tuner/cache_latents.py
index eb042a5..2112241 100644
--- a/src/musubi_tuner/cache_latents.py
+++ b/src/musubi_tuner/cache_latents.py
@@ -300,8 +300,9 @@ def main():
     parser = hv_setup_parser(parser)

     args = parser.parse_args()
-
-    device = args.device if args.device is not None else "cuda" if torch.cuda.is_available() else "cpu"
[...]

PR #813, opened 2025-12-26
Minimum fix to work for MPS backend (for Mac OS)

djtuBIG-MaliceX:
# WHAT IS THIS
As per https://github.com/kohya-ss/musubi-tuner/issues/790, some minor code changes were required to allow seamless usage of musubi-tuner GUI via `uv run --extra gui python src/musubi_tuner/gui/gui.py` and actually succeed with running both pre-training and training steps to use GPU instead of CPU via PyTorch MPS backend.

Tested with M3 Ultra 256GB Mac Studio running MacOS Sequoia 15.7.2.

# Notes
- Mixed precision must be set to **no**
- Block swap not working (currently assumes CUDA (or XPU via PR https://github.com/kohya-ss/musubi-tuner/pull/801) which is not available for Mac OS PyTorch, likely addressed elsewhere)
- FP8 not supported
- For MacOS "Darwin" platform, optimizer will always use `adamw` instead of `adamw8bit` due to missing operation per https://github.com/pytorch/pytorch/issues/141287#issuecomment-3690857294 and https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1402 (though it probably should be disabled for CPU as well for all platforms?)
- For larger training sets, you definitely want to set both `PYTORCH_MPS_LOW_WATERMARK` `PYTORCH_MPS_HIGH_WATERMARK` to non-zero values as currently PyTorch MPS backend has a tendency to run away with RAM hogging otherwise.

### DIFF
--- a/src/musubi_tuner/gui/gui.py
+++ b/src/musubi_tuner/gui/gui.py
@@ -924,7 +924,7 @@ num_repeats = 1
             # Construct the full command string
             # list2cmdline will quote arguments as needed
             inner_cmd_str = subprocess.list2cmdline(inner_cmd)
-            is_windows = (os.name == "nt")
+            is_windows = os.name == "nt"
 
             # Chain commands: Run training -> echo message -> pause >nul (hides default message)
             if is_windows:
@@ -941,7 +941,7 @@ num_repeats = 1
                     flags = subprocess.CREATE_NEW_CONSOLE
                     # Pass explicit 'cmd', '/c', string to ensure proper execution
                     cmd_list = ["cmd", "/c"].extend(cmd_list)
-                
+
                 subprocess.Popen(cmd_list, creationflags=flags, shell=(is_windows is False))
                 if is_windows:
                     return f"Training started in a new window! / 新しいウィンドウで学習が開始されました！\nCommand: {inner_cmd_str}"
@@ -1145,4 +1145,4 @@ num_repeats = 1
 
 if __name__ == "__main__":
     demo = construct_ui()
-    demo.launch(i18n=i18n, server_name="0.0.0.0")
+    demo.launch(i18n=i18n)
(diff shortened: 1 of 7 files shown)


src/musubi_tuner/gui/gui.py
```diff
@@ -824,7 +825,7 @@ def start_training(
                 "--network_dim",
                 str(int(dim)),
                 "--optimizer_type",
-                "adamw8bit",
+                "adamw" if platform.system() == "Darwin" else "adamw8bit",
```
djtuBIG-MaliceX, line 828: I think really the condition should be "If running on Mac OS or otherwise device is CPU" but not sure of a clean way to do this here.

kohya-ss: I think this works well enough. If in the future bitsandbytes becomes able to work with mps, or if it is discovered that there are devices other than mps that bitsandbytes does not work with, I think it would be fine to rewrite it in an appropriate way.

kohya-ss:
Thank you for this PR! Could you please format it in ruff?

src/musubi_tuner/gui/gui.py
```diff
@@ -1131,4 +1145,4 @@ def update_model_info(model):
 
 if __name__ == "__main__":
     demo = construct_ui()
-    demo.launch(i18n=i18n)
+    demo.launch(i18n=i18n, server_name="0.0.0.0")
```
kohya-ss: Is this change required for mps? Since this makes the PC accessible from other than localhost, I think it would be better to not enable it for this GUI, or to make it optional.

src/musubi_tuner/gui/gui.py
djtuBIG-MaliceX: yea that's fine. leave it for now until more people actually use it i guess

src/musubi_tuner/gui/gui.py
djtuBIG-MaliceX: Whoops missed this. Fixed in 6d3e1f4.

djtuBIG-MaliceX:
> Thank you for this PR! Could you please format it in ruff?

Done!

giovanni-amati:
Hi, when is this going to be merge? Thanks

PR #818, opened 2025-12-28
Add all models and parameters to GUI, replace accordions with a tabbed wizard interface

Tophness:
Will likely need a lot of testing and tweaking still, but I thought I'd save you some time.
Incase the screenshots are confusing, I just made this a [plugin](https://github.com/Tophness/wan2gp-training/) for [WAN2GP](https://github.com/deepbeepmeep/Wan2GP), but you can just ignore the bits at the top and bottom. This PR is for the standalone gui.py. The WAN2GP plugin just hooks into musubi-tuner's gui and throws it in a tab.

image

image

image

image


kohya-ss:
Thank you for your pull request and the effort you put into this implementation!

I appreciate the comprehensive features you've added. However, I'd like to keep the GUI in this repository intentionally simple for the following reasons:

1. **Maintenance cost**: Complex GUIs require significant effort to maintain and update
2. **User experience**: Too many options can overwhelm beginners. By limiting settings to the essentials, users can start training immediately without confusion

For users who need advanced configurations, I expect them to use the CLI interface.

Therefore, I'm afraid I cannot merge this PR as-is. I hope you understand.

Alternative suggestion: Would you consider publishing your rich GUI as a separate repository, using Musubi Tuner as a library? Such a tool would be valuable for many users, and I'd be happy to link to your repository in the README as an alternative GUI option.

I hope we can find a way to make your contribution valuable to the community.

Thank you for your understanding.

Tophness:
I totally get that. The reason for using tabs was to make it simpler to navigate because the whole page of stuff was overwhelming for me even with the simpler options. It was much neater and more intuitive when it was organised with tabs initially until I filled it with all the extra stuff.

Perhaps we could get the best of both worlds.
We could limit everything by default and put all advanced options in accordions?
It already kind of is that way on the training tab, I've just expanded some advanced options in the screenshots so you can see what's possible.

As for maintenance, maybe something a bit more modular.
Again, already headed that way (models are defined with dynamic dicts in config_manager.py), but we can move everything to json so it's easy to update as things change (or even get an LLM to update it - can't really go catastrophically wrong and break things if it's just updating json).
Could make this entirely optional too, like if the user is in simple mode (default), it resorts to the structure you have now and doesn't even export the advanced args at all (since they may be invalid/outdated), but if they use advanced options, it's more of an experimental thing that may or may not be updated, and that the community can share simple fixes for by providing the extra json for new arguments or models you've added.

I think there are already a few other rich GUIs out there (haven't used them yet), but they seemed much less intuitive to set up than your one and came with their own musubi-tuner forks that were outdated.
It'd be good to have an official one that's easy for the community to update if it doesn't slow things down.

Tophness:
Antigravity proposed the following json structure and this makes a lot of sense.

```
{
  "models": {
    "Z-Image-Turbo": {
      "resolution": [1024, 1024],
      "vae_subpath": ["vae", "ae.safetensors"],
      ...existing config_manager data...
    },
    ...all other models...
  },
  "training_args": {
    "essential": [
      {"name": "output_name", "default": "my_lora"},
      {"name": "network_dim", "default": 16},
      {"name": "learning_rate", "default": 1e-4},
      {"name": "num_epochs", "default": 16},
      {"name": "save_every_n_epochs", "default": 1},
      {"name": "discrete_flow_shift", "default": 2.0}
    ],
    "advanced": {
      "optimizer": [...optimizer args...],
      "flow_timestep": [...weighting, logit, timestep args...],
      "network_lora": [...alpha, dropout args...],
      "training_flow": [...seed, grad norm, resume args...],
      "metadata": [...title, author, license args...],
      "huggingface": [...repo, token args...]
    }
  },
  "dataset_options": {
    "essential": ["image_directory", "video_directory", "cache_directory"],
    "advanced": ["jsonl_file", "control_directory", "target_frames", "frame_extraction", ...]
  },
  "model_specific_args": {
    "Wan": {
      "essential": ["t5", "dit"],
      "advanced": ["clip", "fp8_scaled", "fp8_t5", "vae_cache_cpu", "one_frame", "offload_inactive_dit", "force_v2_1_time_embedding", "dit_high_noise", "timestep_boundary"]
    },
    ...other models...
  }
}
```

Tophness:
Just implemented those changes and improved the styling a bit. New screenshots:

Initialization tab:
image

Models tab (maybe too many path options available here, let me know if any should be moved to advanced):
image

Dataset editor tab:
image

Dataset editor tab (video dataset added - advanced options collapsed):
image

Training tab:
image

Tools tab (can remove the command line preview, but this is useful for debugging during development, might be useful to have .bat/.sh export options for users to run outside the browser though):
image

kohya-ss:
Thank you for your suggestions! I appreciate you thinking through various approaches like tabs, accordions, and JSON configuration.

However, I'd like to clarify our design philosophy: this GUI is intentionally simple by design. Even with better organization through tabs or accordions, I want to avoid adding more configuration options. This is a product direction, not a technical limitation.

Additionally, the scale of this PR (3,700 lines added) makes it difficult to review and maintain. Regardless of implementation approach, more features inevitably mean higher ongoing maintenance costs.

I'm also not planning to have an "official rich GUI." Making it official would mean committing to support all features continuously, and I need to focus my resources on core development of Musubi Tuner itself.

Since you find existing rich GUIs less intuitive, I encourage you to publish your version as a separate repository. That way, you can freely add and improve features at your own pace, and I can link to it from the README.

Again, this isn't about code quality—your work is excellent. It's simply a matter of repository direction.

I hope you understand.

Tophness:
Well the new version is specifically designed so there would be no need to "commit to support all features continuously" and so the GUI would remain "simple by design". If there's anything in particular you mean by the design still not being simple enough, I can amend it.

By default the idea is for it to work exactly like the original. If it still looks like there's too much in the model or training tabs, those extra options still remaining can be moved to advanced tabs.
If even the collapsed accordions are taking up too much space or seem too overwhelming for the user, there could be an --advanced launch arg or an advanced mode switch at the top so they're completely hidden by default. The experience would be literally visually identical.

And when it comes to maintaining the json, if all you want is for the current basic mode (based on your original GUI) to work with future updates and not have to maintain it, that's currently how it's set up. All the advanced mode args are optional, so if you don't maintain those, the GUI would still be functionally identical to the one you have now, and the outputted args would be identical to those outputted by your original UI.

> the scale of this PR (3,700 lines added) makes it difficult to review and maintain

I can understand how that would seem daunting, although most of that is just boilerplate for things like dataset verification libs, the json loader, (probably unnecessary) css, etc. which never need to change. The initial json will be big once on the first push because it includes all possible models and options, but if you even had to update it, that would just mean going to the basic mode parts and changing the handful of lines in those sections. Could even have a config_basic.json and config_advanced.json if you don't even want to see the rest of it.
The only real code counting towards that in any real way is in gui.py - 1563 lines vs your original 1134.

I'm just putting forward once last time that there seems to be (from my perspective) no foreseeable advantage to maintainability/design philosophy sticking to the original, while potentially a lot to gain with the option of a fully featured UI.
Having to install a separate UI and link it to a separate musubi-tuner install would also make it harder to use/maintain (it imports musubi_tuner.dataset.config_utils, so i would have to keep redundant copies of musubi-tuner code and update them with every change) and for noobs to keep up to date when they have to maintain multiple repos across separate installations, install separate requirements, potentially separate venvs etc.
I think this is why a lot of the external GUIs got abandoned.

I'll make it a separate repo anyway since it seems like that's what you want, but if any of this made you reconsider that we can have both with no consequences, let me know and I'll do whatever it takes to make the default mode as simple as you want.
I wouldn't want you to have to spend time maintaining UI stuff either. I actually thought the json would be easier to update than the hardcoded python.
I'd be happy to maintain the advanced stuff as a co-contributor on a separate branch if you're worried you'd feel obliged to update that too (or there's the option of fetching config_advanced.json remotely and never needing to update the repo at all).
I'm generally monitoring updates to this repo anyway so I wouldn't mind adding new args as they arrive.

kohya-ss:
Thank you for your detailed response and for taking the time to address each of my concerns. I really appreciate your enthusiasm and willingness to maintain this.

I think forking this repository would be the best approach. Here's why:

**Addresses your concerns:** Since you import `musubi_tuner.dataset.config_utils`, forking allows you to pull upstream changes directly. This eliminates the need for redundant code copies that you mentioned. Users would only need to clone your single fork—no multiple repositories to manage.

**Maintenance workflow:** You mentioned you're already monitoring this repo for updates. With a fork, you can simply merge upstream changes when needed, which should fit naturally into that workflow.

**Clear boundaries:** This keeps the base repository simple as intended, while your fork can freely evolve with advanced features. Users who want the rich GUI know exactly where to find it.

I'd be happy to add a link to your fork in the README once it's set up, so users can easily discover it.

Thank you again for your understanding and for your excellent work. I look forward to seeing your GUI project develop!

Tophness:
Well it's already forked at https://github.com/Tophness/musubi-tuner

I don't think a fork is going to be maintainable though. In it's current state, I wouldn't mind adding new args to a json file every couple weeks or so as they arrive so people can optionally try new args, knowing the UI would work and remain up to date regardless, but as a whole fork of musubi-tuner, the UI repo would be out of sync with any other changes to musubi-tuner until I pull upstream changes.
That's basically the state of the other external UIs for sd-scripts/musubi-tuner. There'll be changes to the remote repo you only get if you use the cli until you wait for the ui maintainer to update, and then if they stop maintaining it for even a month or so, the UI becomes horribly out of date.

Btw, after removing a bunch of redundancy, there are now only 319 new lines in this PR. That's including the json file loading functions and everything which isn't maintenance code.
gui.py is now 632 lines compared to your original 1134 lines.
A lot of this is because the UI is generated dynamically by the json config instead of being hardcoded logic, which would have reduced your code significantly. I had left a lot of the original hardcoded logic there even after porting it over, but it was redundant.

kohya-ss:
Thank you for your continued effort, including reducing the line count.

However, I'd like to explain my perspective on the code style. While the refactored version has fewer lines in gui.py, this is achieved through abstraction—dynamic UI generation from JSON, utility functions, etc. The total system complexity (including the config loader, JSON files, etc.) hasn't necessarily decreased.

I personally prefer simple, explicit, and even redundant code that can be understood by reading a single file top to bottom. For example:

```python
# I prefer this - verbose but immediately clear
with gr.Row():
    dit_path = gr.Textbox(label="DiT Path", placeholder="...", max_lines=1)
```

over:

```python
# Harder to trace what UI is actually generated
for arg in arg_loader.get_args():
    ui_components[arg["name"]] = create_component(arg)
```

This isn't about code quality—it's a matter of philosophy. I find the original style easier to maintain because I can see exactly what exists without jumping between files.

Regarding fork maintenance: I understand your concern that forks can fall out of sync. However, if maintaining a fork feels unsustainable, maintaining additional code within this repository would likely face similar challenges over time.

I've given this considerable thought, but I don't think I can merge this PR. I hope you'll consider maintaining your fork for users who prefer a feature-rich GUI. I'm still happy to link to it from the README.

Thank you for understanding.

Tophness:
> I personally prefer simple, explicit, and even redundant code that can be understood by reading a single file top to bottom. For example:
> 
> ```python
> # I prefer this - verbose but immediately clear
> with gr.Row():
>     dit_path = gr.Textbox(label="DiT Path", placeholder="...", max_lines=1)
> ```
> 
> over:
> 
> ```python
> # Harder to trace what UI is actually generated
> for arg in arg_loader.get_args():
>     ui_components[arg["name"]] = create_component(arg)
> ```

Well if you had to update the DiT path (or add one for a new model) it wouldn't be in the code, it would be in the json (which should be much more straight-forward) like
```
"model_specific_args": {
	"HunyuanVideo": {
		"args": [
			{ "name": "hv_dit", "type": "text" }
		]
	}
}
```

Which would create a Textbox gradio component for a DiT path mapped to the hv_dit arg.
hv_dit would have a translation at lbl_hv_dit that says "Hunyuan DiT Path". In your pseudocode example, I think you would normally do this translation step too like `dit_path = gr.Textbox(label=lbl_dit, placeholder="...", max_lines=1)`.

You wouldn't have to be jumping between files or even having to process any hardcoded/abstracted UI code. The config.json (or config_basic.json if the full one is overwhelming) would be where you update everything.
In fact it should be more centralised, because the config.json file generates UI elements that you would otherwise have to update both config_manager.py and gui.py for.
Hell you could even have a common json file for the parameters that musubi-tuner and the UI uses, or even have it scrape them from the list of python args, then the UI stays automatically updated with musubi-tuner with 0 input. I genuinely think there's a way to make this significantly more maintainable while being more functional.

> However, if maintaining a fork feels unsustainable, maintaining additional code within this repository would likely face similar challenges over time.

Well not really, because in it's current state, if there are updates to musubi-tuner like a more efficient internal method that reduces vram or something, nothing would have to update. The UI would automatically take advantage of that the second you push. That's what I love about finally having an official gui included with the repo.
Even if I went away for a year or something and didn't update the json, everything would remain fully functional in basic mode, and if you used advanced mode, the only things that wouldn't work would be new args (which in the worst case scenario, people from the community would probably just comment how to add in PRs for the new features anyway).
If it were a fork, I would have to be constantly monitoring for that push to the official repo and sync that update.
In the worst case scenario where I went away for a year, anyone using the UI would have a musubi-tuner that's a year old with missing models, less efficient memory usage etc. So it would basically be anywhere from outdated to extremely outdated for the same or more amount of input that would otherwise keep it stable and updated at all times.

In it's current state, if there are updates to musubi-tuner that require adding new args, you would just add those new args in the json.
If it were a fork, I would have to manage synchronising to the latest musubi-tuner around pushing updates to the fork, and then if there are any accidental merge conflicts (or real ones - such as an update to musubi_tuner.dataset.config_utils or src/musubi_tuner/gui), sorting those out first.

Fair enough though, it's totally fine if we have a difference of opinion on this. Your concerns are reasonable too.
I don't think I'll be able to maintain an up to date fork with the rate things change in this space if I'm honest about it though. I understand why the external UIs that failed did and can't promise this would be any different.
I should be able to get what I wanted anyway via the WAN2GP plugin because that can manage the musubi-tuner install itself and patch this in, but if you ever change your mind, I'm happy to try to make this standalone UI work for everyone else in a way that doesn't introduce maintenance cost for you, and I can commit to updating the advanced args in the json at a casual pace.

PR #830, opened 2026-01-07
Add save on demand feature

Tophness:
Lately since I've been using runpod and training locally at the same time, I often want to run inference to test out my latest runpod models, but can't while I'm training locally, so I wanted the ability to save on demand by pressing a shortcut key (s) in console rather than having to predefine it in args (can never predict when I'll need it, and saving too much means too much unnecessary writes).
This just reuses the existing save_every_n_steps block.

### DIFF
--- a/src/musubi_tuner/utils/interactive_utils.py
+++ b/src/musubi_tuner/utils/interactive_utils.py
@@ -1,60 +1,84 @@
-import os
 import sys
+import os
+import threading
+import atexit
 import logging
 
 logger = logging.getLogger(__name__)
 
+_save_requested = False
+_input_thread_started = False
+
+def _restore_settings(fd, old_settings):
+    try:
+        import termios
+        termios.tcsetattr(fd, termios.TCSADRAIN, old_settings)
+    except Exception:
+        pass
+
+def _input_listener():
+    global _save_requested
+    
+    try:
+        import tty
+        import termios
+        
+        fd = sys.stdin.fileno()
+
+        if not os.isatty(fd):
+            return
+
+        old_settings = termios.tcgetattr(fd)
+        atexit.register(_restore_settings, fd, old_settings) # restores original terminal settings on exit
+        tty.setcbreak(fd)
+        
+        while True:
+            char = sys.stdin.read(1)
+            
+            if char.lower() == 's':
+                logger.info("\nSave requested via 's' key")
+                _save_requested = True
+
+            if char == '\x03': # restores ctrl+c behaviour
+                _restore_settings(fd, old_settings)
+                import signal
+                os.kill(os.getpid(), signal.SIGINT)
+                break
+                
+    except Exception:
+        pass
+
+def start_listener():
+    global _input_thread_started
+    if _input_thread_started:
+        return
+
+    try:
+        if sys.stdin.isatty():
+            t = threading.Thread(target=_input_listener, daemon=True)
+            t.start()
+            _input_thread_started = True
+    except Exception:
+        pass
+
 def is_save_requested():
-    """
-    Checks if the 's' key has been pressed in the console.
-    This function is non-blocking and should be called periodically.
-    """
+    global _save_requested
     if os.name == 'nt':
         try:
             import msvcrt
-            if msvcrt.kbhit():
-                key = msvcrt.getch()
-                if key.lower() == b's':
-                    logger.info("Save requested by user (keyboard interrupt 's')")
+            while msvcrt.kbhit():
+                if msvcrt.getch().lower() == b's':
+                    logger.info("Save requested by user")
                     return True
-                # consume other keys
-        except Exception:
+        except:
             pass
+        return False
+
     else:
-        try:
-            import fcntl
-            import termios
-            import tty
-
-            fd = sys.stdin.fileno()
-            if not os.isatty(fd):
-                return False
-
-            old_settings = termios.tcgetattr(fd)
-            old_flags = fcntl.fcntl(fd, fcntl.F_GETFL)
-
-            try:
-                fcntl.fcntl(fd, fcntl.F_SETFL, old_flags | os.O_NONBLOCK)
-                tty.setcbreak(fd)
-                try:
-                    key = sys.stdin.read(1)
-                    if key and key.lower() == 's':
-                        logger.info("Save requested by user (keyboard interrupt 's')")
-                        try:
-                            while sys.stdin.read(1):
-                                pass
-                        except IOError:
-                            pass
-                            
-                        return True
-                except IOError:
-                    pass
-
-            finally:
-                termios.tcsetattr(fd, termios.TCSADRAIN, old_settings)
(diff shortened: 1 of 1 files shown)


src/musubi_tuner/hv_train.py
```diff
@@ -1118,8 +1118,20 @@ def remove_model(old_ckpt_name):
...
+                    # check for on-demand save
+                    save_on_demand = False
+                    if accelerator.is_main_process:
+                        if interactive_utils.is_save_requested():
+                            save_on_demand = True
+                    
+                    if accelerator.num_processes > 1:
+                        signal = torch.tensor([1.0 if save_on_demand else 0.0], device=accelerator.device)
+                        signal = accelerator.reduce(signal, reduction="max")
+                        save_on_demand = signal.item() > 0.5
```
kohya-ss, line 1130: Could you explain what this part is intended to do? Does it take into account when a process other than the main process receives `save_on_demand`?

Tophness: Couldn't tell you to be honest, was wondering the same thing. This was Antigravity's handiwork. Seems redundant to me if save_every_n_steps worked without it.
I'm about to start testing it now, and I'll try removing it to see if it still works all the same.

Tophness: The save state function seemed to work all the same without it on my machine.
random_states_0.pkl and scheduler.bin both produced the same MD5 hash with and without it, while everything else changed. Not sure how scientific that is.
I'm only using a single GPU though, and Gemini offered this explanation (browser only, only fed it hv_train.py and interactive_utils.py as context), which makes sense to me. I was wondering whether the other imports were prepared for a save state point that wasn't predetermined, so I did ask Antigravity to investigate before writing any code.

### Is the `accelerator.reduce` block necessary?

**Yes, it is absolutely necessary.** In fact, removing it will crash or freeze your training if you are using more than one GPU.

Here is why:

1.  **The Step Counter is Deterministic:** In your existing code (`save_every_n_steps`), every GPU calculates `global_step` mathematically. If GPU 0 thinks it is step 500, GPU 1 also knows it is step 500. They naturally agree to enter the saving block and hit `accelerator.wait_for_everyone()` at the same time.
2.  **Keyboard Input is Non-Deterministic:** When you press 's', **only the Main Process (Rank 0)** receives that keystroke.
    *   **Without `reduce`:**
        *   **GPU 0:** Detects 's'. Sets `save_on_demand = True`. enters the saving block. Stops and waits at `accelerator.wait_for_everyone()`.
        *   **GPU 1:** Does *not* detect 's'. Sets `save_on_demand = False`. skips the saving block. Continues to the next line of code (trying to train the next step).
    *   **The Result:** Deadlock. GPU 0 is waiting forever for GPU 1 to say "I'm ready to save," but GPU 1 is busy trying to calculate gradients for the next image. The training hangs.

**The `accelerator.reduce` block acts as a communicator.** It takes the "True" signal from GPU 0, broadcasts it to all other GPUs, so everyone agrees to stop and save together.

src/musubi_tuner/utils/interactive_utils.py
```diff
@@ -0,0 +1,47 @@
...
+                # Check if data is available to read
+                if select.select([sys.stdin], [], [], 0)[0]:
+                    key = sys.stdin.read(1)
+                    if key.lower() == 's':
+                        logger.info("Save requested by user (keyboard interrupt 's')")
+                        return True
+            finally:
+                termios.tcsetattr(sys.stdin, termios.TCSADRAIN, old_settings)
+        except Exception:
+            pass
```
Tophness: This appears to not be working.
Windows version works fine, but I tested the linux version on a runpod/ubuntu server via ssh and using docker desktop in a cmd window launched via `docker exec -it musubi-tuner-gpu bash`. Pressing s did nothing.

Not sure where to go from here yet. It should be working, but isn't.
This would be a working alternative, but it's not great:

```
TRIGGER_FILE = "save_trigger"
    if os.path.exists(TRIGGER_FILE):
        logger.info(f"Save requested via file trigger: {TRIGGER_FILE}")
        try:
            os.remove(TRIGGER_FILE) # remove to avoid infinite loop
        except OSError:
            pass
        return True
```

Tophness: I created a test script to confirm the issue existed with a regular python script and nothing else running.
I then updated it with the following code to try a non-blocking method, which got it to work:
```
import sys
import os
import time
import fcntl
import termios
import tty

def test_input():
    print("--- TESTING FCNTL METHOD ---")
    print("Press 's' to trigger. Press 'q' or 'Ctrl+C' to exit.")
    print("--------------------------------")

    fd = sys.stdin.fileno()

    # 1. Validate TTY
    if not os.isatty(fd):
        print("ERROR: Not connected to a TTY. This script must run with 'docker exec -it' or 'docker attach'.")
        return

    # 2. Backup terminal settings
    old_settings = termios.tcgetattr(fd)
    
    # 3. Set File Descriptor to Non-Blocking Mode
    # This ensures sys.stdin.read() throws an error instead of hanging if no key is pressed
    old_flags = fcntl.fcntl(fd, fcntl.F_GETFL)
    fcntl.fcntl(fd, fcntl.F_SETFL, old_flags | os.O_NONBLOCK)

    try:
        # 4. Set Terminal to C-Break mode (No Enter required, but allow Ctrl+C)
        tty.setcbreak(fd)

        counter = 0
        while True:
            try:
                # Try to read 1 byte. 
                # Because of O_NONBLOCK, this will instantly raise IOError if no key is pressed.
                key = sys.stdin.read(1)
                
                if key:
                    if key.lower() == 's':
                        print(f"\n[SUCCESS] 's' detected at count {counter}!")
                        # Clear line for cleanliness
                        sys.stdout.write(f"\rWaiting... {counter}")
                        sys.stdout.flush()
                    elif key.lower() == 'q':
                        print("\nQuitting.")
                        break
                    else:
                        # Echo other keys just to prove it receives input
                        print(f"\nInput detected: {repr(key)}")
                        sys.stdout.write(f"\rWaiting... {counter}")
                        sys.stdout.flush()

            except IOError:
                # This is expected when no key is pressed in non-blocking mode
                pass

            # Visual heartbeat to prove loop is running
            if counter % 10 == 0:
                sys.stdout.write(f"\rWaiting... {counter}")
                sys.stdout.flush()
            
            counter += 1
            time.sleep(0.1)

    except KeyboardInterrupt:
        print("\nCtrl+C detected.")
    finally:
        # 5. RESTORE SETTINGS (Crucial, or your terminal will be broken after exit)
        termios.tcsetattr(fd, termios.TCSADRAIN, old_settings)
        fcntl.fcntl(fd, fcntl.F_SETFL, old_flags)
        print("\nTerminal settings restored.")

if __name__ == "__main__":
    test_input()
```

However, after incorporating this non-blocking fix into interactive_utils.py, it's still not working.

Tophness: I've fixed it with a somewhat ugly solution: Listening in a separate thread.

PR #841, opened 2026-01-11
Flux.2-dev training

christopher5106:
Here is a first draft. Please review and I'll update consequently.

Here are a few notes:
- I've not used any prompt upsampling during training. 
- I'm also not using Differential Output Preservation (DOP) but it would be not compatible with caching
- I'm using text encoder in bfloat16
- I didn't check training schedule, I'm reusing the timesteps provided by the framework... The get_schedule method for inference comes directly from BFL repo.
- since BFL autoencoder already packs/unpacks the latents during encoding/decoding, the training is performed on the packed loss (does not change anything in the mean).

I'm going to test it this week more.

### DIFF
--- a/src/musubi_tuner/flux_2/flux2_utils.py
+++ b/src/musubi_tuner/flux_2/flux2_utils.py
@@ -871,15 +871,6 @@ class Qwen3Embedder(nn.Module):
         out = torch.stack([output.hidden_states[k] for k in OUTPUT_LAYERS_QWEN3], dim=1)
         return rearrange(out, "b c l d -> b l (c d)")
 
-    def test_txt(self, txt: str) -> bool:
-        raise NotImplementedError("Qwen3Embedder does not support text testing")
-
-    def test_image(self, image) -> bool:
-        raise NotImplementedError("Qwen3Embedder does not support image testing")
-
-    def upsample_prompt(self, txt: list[str], img=None, **kwargs) -> list[str]:
-        raise NotImplementedError("Qwen3Embedder does not support upsampling")
-
 
 def load_textembedder(
     model_version: str,


FurkanGozukara:
awesome looking forward to this @kohya-ss and thanks @christopher5106

n0valis:
Good luck. I tried do adapt Kijays comfyui-fluxtrainer (based on Kohya) to Flux2. Got it running on a 5090 but it was slow and absolutely unstable (hard crashes). So I gave up. I'm excited about your project.

kohya-ss:
Thank you for your efforts! Can I assume it's ready for review and merging?

christopher5106:
I'm finishing the `flux_2_generate_image.py` today and yes, it should be ready for merge.
Training results are good, in particular for details.

n0valis:
> I'm finishing the `flux_2_generate_image.py` today and yes, it should be ready for merge. Training results are good, in particular for details.

On which hardware are you testing? And how long does it take?

kohya-ss:
> I'm finishing the `flux_2_generate_image.py` today and yes, it should be ready for merge. Training results are good, in particular for details.

Thank you so much! I will review and merge sooner😄

christopher5106:
> > I'm finishing the `flux_2_generate_image.py` today and yes, it should be ready for merge. Training results are good, in particular for details.
> 
> On which hardware are you testing? And how long does it take?

I train for 2 hours on H100, 3000 steps, 10 images, 20 repeats

christopher5106:
> > I'm finishing the `flux_2_generate_image.py` today and yes, it should be ready for merge. Training results are good, in particular for details.
> 
> Thank you so much! I will review and merge sooner😄

Mostly what remains:
- flow shift values to check
- prompt upsampling could be added to generation (I can do it in another PR)
-  fp8 quantization

christopher5106:
I'm done with image generation script.

To generate on H100, since VAE is quite heavy, I had to add a move of the transformer to CPU
https://github.com/scenario-labs/musubi-tuner/blob/flux2/src/musubi_tuner/flux_2_generate_image.py#L645
Could you please check if it's compatible with interactive mode ? (if you move it back to device at next generation for example)

christopher5106:
My implementation was inspired mainly by your Flux Kontext training implementation. But to make it compatible with multiple reference images, I also took inspiration from your Qwen Edit training implementation. For the core part, I copy-pasted BFL original code, which is in Diffusers format for the models, and it makes it easy to follow. I kept their way to pack/unpack in the encode and decode forward pass, not outside, so the latents on disk have a different shape than for Flux Kontext.

n0valis:
> My implementation was inspired mainly by your Flux Kontext training implementation. But to make it compatible with multiple reference images, I also took inspiration from your Qwen Edit training implementation. For the core part, I copy-pasted BFL original code, which is in Diffusers format for the models, and it makes it easy to follow. I kept their way to pack/unpack in the encode and decode forward pass, not outside, so the latents on disk have a different shape than for Flux Kontext.

Do you think there is any chance to do this on a 5090 /w blockswapping on cuda1(16GB) and CPU(96GB RAM)?

christopher5106:
Block swapping is implemented the same way as for Flux Kontext but I didn't test it, you should try.

n0valis:
> Block swapping is implemented the same way as for Flux Kontext but I didn't test it, you should try.

I will, many thanks for your great work.

christopher5106:
By the way, you can choose to add or not some control images, depending what kind of training you want to perform. I realize that you need a lot more memory to train with control images.

n0valis:
It is very difficult to handle this model without sufficient memory. I have experienced this with my adaption where I had to swap 40 blocks and the blockswapping took so long that the GPU ran idle between 1-5% load at all times. So unless you cooked up some magic soup I don't see it happening on less than 64GB VRAM.

n0valis:
Is it possible/makes sense to train with a flux2-dev-fp8_scaled model?

christopher5106:
Yes I think so.
I'm also adding training of Flux 2. Klein 4B and 8B, I modified the code consequently

kohya-ss:
Cache generation, LoRA training, and image generation all worked without any issues. Great!

> Could you please check if it's compatible with interactive mode ? (if you move it back to device at next generation for example)

Interactive mode works fine.

> My implementation was inspired mainly by your Flux Kontext training implementation. But to make it compatible with multiple reference images, I also took inspiration from your Qwen Edit training implementation. For the core part, I copy-pasted BFL original code, which is in Diffusers format for the models, and it makes it easy to follow. I kept their way to pack/unpack in the encode and decode forward pass, not outside, so the latents on disk have a different shape than for Flux Kontext.

I think this is the right choice of implementation.

I'd like to finish the Klein implementation today, so I'll review and merge this as soon as I have time.
If there's any Klein implementation left, I'd like to implement that as well as add fp8 scaling support.

christopher5106:
I've added Klein in my last commit, did you see ?

kohya-ss:
> I've added Klein in my last commit, did you see ?

Sorry, I didn't check. That's great! It looks like I have very little work left to do. Thank you again!

christopher5106:
[here](https://github.com/scenario-labs/musubi-tuner/blob/flux2/docs/flux_2.md#specifying-model-version--%E3%83%A2%E3%83%87%E3%83%AB%E3%83%90%E3%83%BC%E3%82%B8%E3%83%A7%E3%83%B3%E3%81%AE%E6%8C%87%E5%AE%9A)
But I didn't check, I just checked that Flux2 dev still trains. Remains to add the checkpoints to download, + modifiy the Qwen 3 offline loading 
https://github.com/scenario-labs/musubi-tuner/blob/flux2/src/musubi_tuner/flux_2/flux2_utils.py#L793 
I modified the way to package the text encoder and tokenizer by packaging them together in a Embedder class, to align it with BFL last implementation

kohya-ss:
> I modified the way to package the text encoder and tokenizer by packaging them together in a Embedder class, to align it with BFL last implementation

That's good. I'll consider refactoring it when a model using Qwen3-VL comes out in the future.

kohya-ss:
Thank you very much for the great work!

n0valis:
> Block swapping is implemented the same way as for Flux Kontext but I didn't test it, you should try.

It crashed for me with --blocks_to_swap 12

Here is what Gemini  found regarding the block swapping:
 CRITICAL FINDING:
  submit_move_blocks_forward is used in flux_models.py (Flux 1), hunyuan_model.py, etc.
  IT IS NOT USED IN `flux2_models.py`!

  The Flux2.forward method in projects/musubi-tuner/src/musubi_tuner/flux_2/flux2_models.py is MISSING the block swapping logic entirely. It initializes the offloader but never calls it during the forward pass.

  This is a bug in the Musubi implementation for Flux 2. It sets up the infrastructure for swapping but forgot to hook it into the inference/training loop.

  Fix:
  I must modify flux2_models.py to call submit_move_blocks_forward inside the forward loop.

christopher5106:
Please try and tell me if you see other issues https://github.com/kohya-ss/musubi-tuner/pull/862

src/musubi_tuner/flux_2/flux2_utils.py
```diff
@@ -0,0 +1,889 @@
...
+    return x_list
+
+
+def encode_image_refs(ae, img_ctx: list[Image.Image]):
+    scale = 10
+
+    if len(img_ctx) > 1:
+        limit_pixels = 1024**2
+    elif len(img_ctx) == 1:
+        limit_pixels = 2024**2
```
kohya-ss, line 143: Setting `limit_pixels` to `2024**2` doesn't seem to exist in Diffusers' `pipeline_flux2_klein.py`. Where does this code come from?

christopher5106: Yes I saw. But originally in BFL code that is an extract of their code base, it is like that https://github.com/black-forest-labs/flux2/blob/b56ac61450f56ea7d32374c2fa54e77a262067f6/src/flux2/sampling.py#L56

christopher5106: It is also not available in pipeline_flux2.py as well, no? But one of the benefits of Flux2 is this capability to edit images in higher resolution...

kohya-ss: Thank you for clarification! I just noticed the existence of an official flux2 repository.

The [introduction page for klein on BFL](https://bfl.ai/models/flux-2-klein) only have a link to HuggingFace weights and FLUX.1 repository, and the former instructs us to use Diffusers, so I thought Diffusers was the official.

So it seems that the Diffusers implementation is insufficient...

christopher5106: You could add a note somewhere for those who are interested to train in low VRAM mode, to resize their control images in low resolution, and limit their number of control images. 

Shouldn't we ask Diffusers to increase their limit and align their implementation?

kohya-ss: I would like to refactor the scripts so that resizing of control images is done on the dataset side rather than the cache script side. As with the existing Qwen-Image-Edit, I will allow the resolution of control images to be explicitly specified. This will allow users to specify 1024, 2024, or any other resolution regardless of the number of control images.

Diffusers seems to have many other deficiencies, so it would be best to wait until all implementation is complete before proceeding.

src/musubi_tuner/flux_2_generate_image.py
```diff
@@ -0,0 +1,1169 @@
...
+        1,
+        128,
+        packed_latent_height,
+        packed_latent_width,
+        dtype=noise_dtype,
+        generator=seed_g,
+        device="cpu",
+    ).to(device, dtype=torch.bfloat16)
+    x, x_ids = flux2_utils.batched_prc_img(noise)
+    # TODO upsampling here
```
kohya-ss, line 619: What exactly is this upsampling process? Please let me know if anything needs to be done.

christopher5106: There is a local prompt upsampling using Mistral  3 (for both Flux 2. dev and klein) to generate better images.
https://github.com/black-forest-labs/flux2/blob/main/scripts/cli.py#L553 or https://github.com/huggingface/diffusers/blob/1eb40c6dbdf51f54514febf2fdea827a5414c494/src/diffusers/pipelines/flux2/pipeline_flux2.py#L867

kohya-ss: Thank you for clarification. Users can pre-upsample the prompt with any LLM if necessary, so this step may be omitted.

PR #843, opened 2026-01-12
Support Z-image-omni-base

sdbds:
https://github.com/huggingface/diffusers/commit/f6b6a7181eb44f0120b29cd897c129275f366c2a
Code Reference↑


kohya-ss:
Thank you for this! This is great!

It looks like we can test the code without the omni-base weights. I will review this sooner. I think the code might be cleaned up a bit after the merge.

sdbds:
> Thank you for this! This is great!
> 
> It looks like we can test the code without the omni-base weights. I will review this sooner. I think the code might be cleaned up a bit after the merge.

In fact, omnibase mainly adds a siglip2 layer for input, and there are almost no changes in the rest, I think this part can be handled by referring to the qwen-edit method.
However, considering that joint training can be done, it is possible to decide whether to use the siglip2 model cache latents based on the presence of control_directer in the dataset

kohya-ss:
The pull request seems to work fine for the existing Z-Image-Turbo training, so I will test and merge this as soon as the base weights are released.

kohya-ss:
Hmm Z-Image-Base weight is released, but Omni is not yet...

sdbds:
> Hmm Z-Image-Base weight is released, but Omni is not yet...

Yes, it might take a while longer.

BenDes21:
> > Hmm Z-Image-Base weight is released, but Omni is not yet...
> 
> Yes, it might take a while longer.

Hi, according to your latest twitter post, seems the bf16 model is " broken " for training ?

BenDes21:
Saw see on reddit today : 


_The author, "None-南," reports that despite the community spending significant money (tens of thousands in compute costs) and countless hours tuning parameters for the new Z-Image model, training results were consistently poor. Users experienced issues such as grey images, structural collapse, and instability (oscillating between overfitting and non-convergence).

The Root Cause: The "Ancient Bug"
After deep analysis and log auditing with the Z-Image team, the culprit was identified as the bitsandbytes AdamW8bit optimizer.

Z-Image is a DiT model with a single-stream architecture that requires High Dynamic Range (HDR) precision.

The AdamW8bit optimizer relies on an outdated Uint8 format. This format has a too narrow range for Z-Image's needs, causing minute gradients to be truncated or zeroed out during training. Essentially, the model was "slacking off" and not learning.

The Solution: Switching to FP8
The author suggests abandoning the 8bit optimizer entirely and has released a custom-wrapped FP8 optimizer (based on native PyTorch support).

Results: Switching to FP8 resulted in approximately 40% faster convergence, eliminated strange noise, improved composition stability, and ensured compute cost translated directly into model quality.

Additional Training Tips from the Author:

Rank: Needs to be 64+; Alpha should typically be half of the Rank.

Captioning: Avoid being too brief or writing overly complex "essays." Use moderate, concise, and accurate descriptions.

Steps: DiT models (like Z-Image) have high internal precision requirements; aim for 1000+ steps (Repeat * Epoch).

Loss: A value between 0.2 and 0.25 typically indicates good performance.

The author has provided the code and configuration demo on GitHub for users to implement immediately._

https://github.com/None9527/None_Z-image-Turbo_trainer/blob/omni/src/zimage_trainer/optimizers/adamw_fp8.py

No idea if its true or not


+ also other informations : 

https://www.reddit.com/r/StableDiffusion/comments/1qwc4t0/thoughts_and_solutions_on_zimage_training_issues/

PR #864, opened 2026-01-22
Fix Flux2 block swap, Flux2.klein training+download doc

christopher5106:
- Fixing https://github.com/kohya-ss/musubi-tuner/pull/841#issuecomment-3780058126 about block swap in Flux2.dev training
- Checkpoint downloads for Flux2.Klein
- Fix for Flux2.klein training+inference

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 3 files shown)


kohya-ss:
Thank you for update!

PR #868, opened 2026-01-24
feat: Sliced SDPA for FLUX.2 Latent Encoding

xzuyn:
Add `--vae_slice_size` option to allow for encoding larger latents without OOM. My 24GB GPU would OOM doing 2048x2048 buckets normally, but using `--vae_slice_size 1024` encodes them no problem in ~2s/it. 2560x2560 buckets work too. 3072x3072 starts to OOM at a different part though, even with `--vae_slice_size 512`.

By default no slicing is used.

### DIFF
--- a/src/musubi_tuner/flux_2/flux2_models.py
+++ b/src/musubi_tuner/flux_2/flux2_models.py
@@ -83,9 +83,10 @@ def swish(x: Tensor) -> Tensor:
 
 
 class AttnBlock(nn.Module):
-    def __init__(self, in_channels: int):
+    def __init__(self, in_channels: int, slice_size: int = None):
         super().__init__()
         self.in_channels = in_channels
+        self.slice_size = slice_size
 
         self.norm = nn.GroupNorm(num_groups=32, num_channels=in_channels, eps=1e-6, affine=True)
 
@@ -104,9 +105,25 @@ class AttnBlock(nn.Module):
         q = rearrange(q, "b c h w -> b 1 (h w) c").contiguous()
         k = rearrange(k, "b c h w -> b 1 (h w) c").contiguous()
         v = rearrange(v, "b c h w -> b 1 (h w) c").contiguous()
-        h_ = nn.functional.scaled_dot_product_attention(q, k, v)
+
+        if self.slice_size is not None:
+            h_ = self.sliced_sdpa(q, k, v)
+        else:
+            h_ = nn.functional.scaled_dot_product_attention(q, k, v)
 
         return rearrange(h_, "b 1 (h w) c -> b c h w", h=h, w=w, c=c, b=b)
+
+    def sliced_sdpa(self, q: Tensor, k: Tensor, v: Tensor) -> Tensor:
+        seq_len = q.shape[2]
+        slice_size = min(self.slice_size, seq_len)
+        output = torch.zeros_like(q)
+        for i in range(0, seq_len, slice_size):
+            end_idx = min(i + slice_size, seq_len)
+            q_slice = q[:, :, i:end_idx, :]
+            attn_slice = nn.functional.scaled_dot_product_attention(q_slice, k, v)
+            output[:, :, i:end_idx, :] = attn_slice
+
+        return output
 
     def forward(self, x: Tensor) -> Tensor:
         return x + self.proj_out(self.attention(x))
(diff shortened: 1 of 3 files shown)


kohya-ss:
Thank you for submitting a PR! This is really useful.

Since vae slicing often refers to slicing an image or latent in the batch element directions, we will change the option name after merging to make it clear that it is attention slice. Thank you for your understanding.

kohya-ss:
After further investigation, it appears that the memory peak is in the high-resolution part of the AE encoder/decoder. Therefore, unfortunately, this PR does not seem to be a fundamental solution. 

We will consider further memory saving measures, but H/W tiling may be necessary.

Please use `--vae_dtype bfloat16` first for memory saving.

xzuyn:
> Since vae slicing often refers to slicing an image or latent in the batch element directions, we will change the option name after merging to make it clear that it is attention slice. Thank you for your understanding.

That's fair. I wasn't fully sure what the correct name for it would be, so I just went with slicing.

> After further investigation, it appears that the memory peak is in the high-resolution part of the AE encoder/decoder. Therefore, unfortunately, this PR does not seem to be a fundamental solution.

Yea, it's not a fix to the entire OOM issue, but it lets me go up to 2560x buckets which is a nice improvement over not even being able to do 2048x buckets.

3072x buckets cause a different OOM, so they would require some other change. IMO this is a fairly small change to add to reduce the peak just enough to get a bit higher resolution when encoding latents. It wasn't meant to solve all OOM problems. If there is a better fix later on, this PR could easily be reverted in favour of it.

> Please use `--vae_dtype bfloat16` first for memory saving.

I've only been using BF16 VAE mode, so my mentions of OOM are with that.

xzuyn:
Some changes could be included that reduce memory
- replace `swish` with `torch.nn.functional.silu` in `ResnetBlock.forward`, `Encoder.forward`, and `Decoder.forward`
- replace `x + h` with `h += x` in `ResnetBlock.forward`
- update `h` inplace in `Encoder.forward` instead of using `hs` list since its always using `hs[-1]`
- add `del z` to `Decoder.forward` to free memory sooner

Also if we need the VAE eventually for other loss types like SSIM, adding checkpointing to `ResnetBlock.forward` works well.

```diff
--- a/src/musubi_tuner/flux_2/flux2_models.py
+++ b/src/musubi_tuner/flux_2/flux2_models.py
@@ -78,10 +78,6 @@
     z_channels: int = 32
 
 
-def swish(x: Tensor) -> Tensor:
-    return x * torch.sigmoid(x)
-
-
 class AttnBlock(nn.Module):
     def __init__(self, in_channels: int):
         super().__init__()
@@ -129,17 +125,19 @@
     def forward(self, x):
         h = x
         h = self.norm1(h)
-        h = swish(h)
+        h = torch.nn.functional.silu(h)
         h = self.conv1(h)
 
         h = self.norm2(h)
-        h = swish(h)
+        h = torch.nn.functional.silu(h)
         h = self.conv2(h)
 
         if self.in_channels != self.out_channels:
             x = self.nin_shortcut(x)
 
-        return x + h
+        h += x
+
+        return h
 
 
 class Downsample(nn.Module):
@@ -218,25 +216,23 @@
         self.conv_out = nn.Conv2d(block_in, 2 * z_channels, kernel_size=3, stride=1, padding=1)
 
     def forward(self, x: Tensor) -> Tensor:
-        # downsampling
-        hs = [self.conv_in(x)]
+        h = self.conv_in(x)
+        del x
         for i_level in range(self.num_resolutions):
             for i_block in range(self.num_res_blocks):
-                h = self.down[i_level].block[i_block](hs[-1])
+                h = self.down[i_level].block[i_block](h)
                 if len(self.down[i_level].attn) > 0:
                     h = self.down[i_level].attn[i_block](h)
-                hs.append(h)
             if i_level != self.num_resolutions - 1:
-                hs.append(self.down[i_level].downsample(hs[-1]))
+                h = self.down[i_level].downsample(h)
 
         # middle
-        h = hs[-1]
         h = self.mid.block_1(h)
         h = self.mid.attn_1(h)
         h = self.mid.block_2(h)
         # end
         h = self.norm_out(h)
-        h = swish(h)
+        h = torch.nn.functional.silu(h)
         h = self.conv_out(h)
         h = self.quant_conv(h)
         return h
@@ -305,6 +301,7 @@
 
         # z to block_in
         h = self.conv_in(z)
+        del z
 
         # middle
         h = self.mid.block_1(h)
@@ -324,7 +321,7 @@
 
         # end
         h = self.norm_out(h)
-        h = swish(h)
+        h = torch.nn.functional.silu(h)
         h = self.conv_out(h)
         return h
```

PR #869, opened 2026-01-24
Flux.2, Flux.1 Kontext: Add support for non-default text embedding length

recris:
This allows the default length of 512 to be changed to something shorter, when training captions don't require that many tokens. This gives a reasonable boost to training speed.

Added a new parameter `--tokenizer_max_length` to `flux_2_cache_text_encoder_outputs.py` script.

Average step time on a Radeon 7900XTX card, after caching using `--tokenizer_max_length 128`, and trained with batch size = 2:
- Before: **13.9 s/it**
- After: **11.5 s/it**

Also cached embedding files go from 12 Mb to 3 Mb, which saves quite some disk space in larger datasets.

I have only tested this with Klein 9B, so far. This also requires cached text embeddings to be re-created.

TODO: implement the same logic for other models.


kohya-ss:
Thank you, this is great!

Would it be possible to add documentation and format the code in ruff?

rockerBOO:
Maybe it makes sense for the default to be much larger. No need to limit it to the arbitrary 512 compared to the supported 128k context length on Mistral. Comfy is doing 2512 for 4B and 4092 for 9B Klein models which might give some  context of how long might be appropriate as a default. I think the downsides of it clipping the caption at 512 might not be enough correlation to other implementations for it to a good default. 

*Edit* im also realizing that we are padding the embeddings so they all match in size. So maybe it makes sense for the caching to warn/error if they go over the token limit. If this is the case it could be done separately, and not change the default  embedding length.

kohya-ss:
FLUX.2's fixed text embedding length is quite strange (or outdated).

If FLUX.2 models could handle text embeddings of arbitrary length, using attention masks might be one way to support arbitrary lengths.

In terms of implementation, it's not that difficult, as we just need to treat it like any other architecture. The default should be padding/trimming to 512, with variable lengths will only be enabled when specified.

recris:
@kohya-ss  added the same option for Flux Kontext - I think this should be good to go.

PR #878, opened 2026-01-28
fix: Update CFG handling in sample generation for Z-Image

kohya-ss:
Also update the documentation.


niceguy4:
Thank you for the super fast update.

niceguy4:
@kohya-ss Does this line need to be changed?

ZIMAGE_ID = "Tongyi-MAI/Z-Image-Turbo"

src/musubi_tuner/zimage/zimage_utils.py

kohya-ss:
> @kohya-ss Does this line need to be changed?
> 
> ZIMAGE_ID = "Tongyi-MAI/Z-Image-Turbo"
> 
> src/musubi_tuner/zimage/zimage_utils.py

Thank you! Z-Image-Turbo and Z-Image use the same configuration for the tokenizer, so this is fine, but I'll update it for consistency.

PR #879, opened 2026-01-28
doc: update link to caption_images_by_qwen_vl.py

obsxrver:
Doc: Updated the link to use the absolute path to src/musubi_tuner/caption_images_by_qwen_vl.py instead of ./


kohya-ss:
Thank you for this!

PR #913, opened 2026-03-09
Self-Flow

rockerBOO:
Screenshot 2026-03-09 at 15-40-57 Self-Supervised Flow Matching for Scalable Multi-Modal Synthesis - self_flow pdf

https://bfl.ai/research/self-flow
[Self-Flow paper PDF](https://cdn.sanity.io/files/2gpum2i6/production/3264c9e34dfcbbf57e57ca7efee0a3e83d311a83.pdf)

**Note: Only Flux2 network training is supported in this code and will probably break other trainings.**

I was reading the paper and realized it could be done on any flow matching model and in fine tuning. To test this I implemented the Self-Flow process for Flux2 network training. 

- Mask ratio: we use teacher noise patches in the student to help "teach" it about lower noise. 0.25 meaning 25% are using the teacher patches/tokens
- Gamma: how much the representation loss is scaled. `loss = L_gen + gamma * L_rep`
- Student/Teacher feature to extract from this layer in the model. 

Flux2 Klein 4B has 25 layers (double and single blocks). 0.3D for student layer, 0.7D for teacher layer.  Where D is number of layers. So 8 (7.5) for student and 18 (17.5) for teacher.

Flux2 Klein 4B config
```toml
self_flow = true
self_flow_gamma = 0.8 
self_flow_gamma_warmup_steps = 100 # experimental to help shorter runs from misalignment in earlier training when representation project and teacher EMA are building out. Check EMA drift graphs for when you might feel an appropriate signal is available.
mask_ratio = 0.25 
ema_decay = 0.999 # 0.999 is fast enough for lower number of steps like 1000, if doing 10k or 100k steps you might want to use 0.9999 which is from the paper. 

student_feature_layer = 8
teacher_feature_layer = 20

# NOTE: Do not use these settings. 
# Experimental distortion of mismatching timesteps between student teacher
# These settings you should not use, it introduces a mismatch but the effect can be artistically interesting 
# self_flow_teacher_coupling_prob = 0.8
# self_flow_teacher_coupling_decay = "rex"
# self_flow_teacher_coupling_decay_steps = 2000
# self_flow_teacher_mismatch_ratio = 0.4
```

```
--self_flow --self_flow_gamma 0.8 --self_flow_gamma_warmup_steps 100 --mask_ratio 0.25 --student_feature_layer 8 --teacher_feature_layer 18
```

Also in the paper they did ImageNet training with Logit Normal distribution. This or flux2_shift should be appropriate. 
```
weighting_scheme = "logit_normal"
logit_mean = 0.0
logit_std = 1.0
```

*I do not think this is the best way to implement this as it requires changing all the network trainers to accept/return new parameters. As well as the model needs to be able to expose some features inside the blocks on the model. *

I am open to feedback about how to integrate this code. 

Issues:

- Batch size > 1 not working
- Different models/trainers not updated to support the features flow

Related papers:
https://arxiv.org/abs/2410.06940
https://arxiv.org/abs/2505.02831


rockerBOO:
Now it will output 3 files instead of 1.

- Student LoRA (the typical file exported)
- Teacher LoRA (ends in `_ema.safetensors`)
- Projection model (ends in `_proj.safetensors`)

You should be using the teacher version which will be much more stable. The performance of the teacher will depend on the EMA decay the most. In the graphs there will be a teacher divergence graph showing when it stablizes or continues to grow more distant. Changing the EMA decay to a smaller value allows the model to align to the student more closely in a shorter amount of steps. 0.999 is a good spot it seems for shorter runs (around 1k). 

Added support for loading the student, teacher EMA and projection to do more steps on these models. 

Fixed mismatches of the timesteps and applied noise in the student training. 

 I also added back a configuration option to have this mismatch happen on purpose. This creates very odd looking images but I am finding the effect kinda cool. Also since the flow is being mismatched it causes the model to get more details from the lower noise version, and can help the model learn some of these details better. The self flow naturally has some of this but the timesteps being properly distributed prevents the flow from being too distorted.  

## Examples of the distortion effect is having
||||
|-|-|-|
|c-f2k4b-2026-03-30_00178_|c-f2k4b-2026-03-29_00278_|c-f2k4b-2026-03-29_00248_|

I will try to put together a set of self-flow and not self-flow examples here soon.

kohya-ss:
Hi @rockerBOO,

Thank you for this PR — Self-Flow is a really interesting technique and I appreciate you taking the time to implement and iterate on it.

I'd like to share a situation and get your input. I've started a medium-scale refactoring of the training script class structure on the [dev branch](https://github.com/kohya-ss/musubi-tuner/tree/dev). It will touch `hv_train_network.py` substantially and would likely conflict heavily with this PR as it currently stands.

I noticed you wrote:

> I do not think this is the best way to implement this as it requires changing all the network trainers to accept/return new parameters.

I share that concern, and I think there may be an approach that addresses both issues:

What do you think about extracting the Self-Flow logic into a dedicated training script — for example `flux_2_train_network_self_flow.py` — that subclasses or composes the base trainer, instead of modifying the shared training path? Benefits:

- Minimizes the conflict surface with the refactoring
- Keeps the Self-Flow code isolated, which also makes review easier
- After the refactoring lands on `dev`, rebasing would mostly be "update the subclass to match the new base class" rather than re-applying scattered changes throughout the trainer

Could you also share your current plans for this PR — rough timeline, remaining blockers, and how you'd like to proceed? That will help me sequence the refactoring work appropriately.

If you'd prefer to keep the current structure and the PR is close to ready, another option is to finalize and merge this PR first, then do the refactoring on top. I'm flexible on the direction — mainly I want to avoid you having to redo a lot of work because of timing.

Thanks again for the contribution!

rockerBOO:
I am trying to work through a refactor of this code to make it more appropriate and limit the amount of injection of self-flow into the standard training process. I have worked through how to refactor out the Self-Flow into SelfFlowStrategy which would work like we are doing with latent/text encoders. 

Second is working on using `register_forward_hook` to capture things in the flow models, so we wouldn't need to modify the base models. 

Biggest abstraction in NetworkTrainer would be how we calculate the loss, which I have in `compute_loss()` which might be worth having separately. I was also considering making a "Mixin" that a user could make their own extensions. This would allow these things like SelfFlow without changing the base trainer code. 

For example this is the abstraction, with kwargs to allow me to extend it:

```python
    def compute_loss(
        self, pred: torch.Tensor, target: torch.Tensor, weighting: float | None, **kwargs
    ) -> tuple[torch.Tensor, dict[str, Any]]:
        """Returns loss and loss metrics"""
        loss = torch.nn.functional.mse_loss(pred, target, reduction="none")

        if weighting is not None:
            loss = loss * weighting
        return loss, {}
```

```python
class SelfFlowMixin:
    def compute_loss(self, ...):

class SelfFlowNetworkTrainer(NetworkTrainer, SelfFlowMixin):
    ...
```

I was wanting to propose some other changes that would allow this sort of extension of musubi tuner to allow these custom things more easily without modifying the base code. 

Some feedback would be splitting out `get_noisy_model_input_and_timesteps` to allow getting timesteps and noisy latents separately. Then i can have it return the timesteps I want, and apply the noisy results.

Also with `call_dit` I am having to modify how this works specifically. Having the ability to compose things or limiting what this type of call is doing outside calling the DiT model. So we are doing patching inside but that could be abstracted into it's own method and then I can modify those patches more easily without having to rewrite `call_dit` to apply my own patches. 

Also with `call_dit` I am needing to send out `features` from the DiT result. This would require changing the output from `call_dit` to add features. 

```python
def call_dit(self, ...):
    return pred, target, features
```

But this means we would need to modify the call sites for each to support this. Otherwise I would need to hack in storing the features in the object (`self.dit_features_capture`) and then in my extension I could capture read that in the compute_loss function. 

Or we make the return a little more flexible with something like.

```python
@dataclass
class DiTOutput:
    pred: torch.Tensor
    target: torch.Tensor
    features: torch.Tensor | None = None

def call_dit(self, ...):
    return DiTOutput(pred=pred, target=target, features=features)
```

Here is the assessment and plan that goes over it in more detail without these refactors: 

[2026-04-23-self-flow-review-assessment.md](https://github.com/user-attachments/files/27057073/2026-04-23-self-flow-review-assessment.md)
[2026-04-23-self-flow-refactor-plan.md](https://github.com/user-attachments/files/27057072/2026-04-23-self-flow-refactor-plan.md)

I could try making self flow, more like how you mentioned, in a specific external implementation with some of those refactors. I will pause on the above plan to see how to proceed based on your feedback.

kohya-ss:
I apologize for the delay. I'm thinking about how to proceed.

rockerBOO:
No worries @kohya-ss. 

I've been working on how the abstraction might look. I got the feature abstraction separated cleanly, and then most of the self-flow is abstracted but the following.

```python
                    if args.self_flow:
                        assert rep_proj is not None
                        assert ema_lora_state is not None
                        loss = self._self_flow_step(
                            args,
                            accelerator,
                            transformer,
                            network,
                            latents,
                            batch,
                            noise,
                            noise_scheduler,
                            dit_dtype,
                            network_dtype,
                            vae,
                            rep_proj,
                            ema_lora_state,
                            global_step=global_step,
                        )
                    else:
                        # --- Vanilla flow matching ---
                        noisy_model_input, timesteps = self.get_noisy_model_input_and_timesteps(
                            args, noise, latents, batch["timesteps"], noise_scheduler, accelerator.device, dit_dtype
                        )

                        weighting = compute_loss_weighting_for_sd3(
                            args.weighting_scheme, noise_scheduler, timesteps, accelerator.device, dit_dtype
                        )

                        output = self.call_dit(
                            args, accelerator, transformer, latents, batch, noise, noisy_model_input, timesteps, network_dtype
                        )
                        model_pred = output.pred
                        target = output.target
                        loss = torch.nn.functional.mse_loss(model_pred.to(network_dtype), target, reduction="none")

                        if weighting is not None:
                            loss = loss * weighting

                        loss = loss.mean()
```

This is similar to the process_batch abstraction in sd-scripts. Then the extending class can rewrite it. like I am doing the `_self_flow_step()` 

```python
def process_batch(
        self,
        args: argparse.Namespace,
        accelerator: Accelerator,
        transformer: torch.nn.Module,
        network: torch.nn.Module,
        latents: torch.Tensor,
        batch: dict,
        noise: torch.Tensor,
        noise_scheduler,
        dit_dtype: torch.dtype,
        network_dtype: torch.dtype,
        vae: torch.nn.Module,
        global_step: int = 0,
):
```

---

Then the logging should have some hook to apply the experiment logging.

```
def on_step_logs(self, args, metrics)
```

Then for sampling we'd need to be able to extend how that works, in this case I need to use the teacher network instead for sampling. 

```python
def on_sample(self, args, network)
```

And lastly is saving, where I need to save the teacher EMA and representation projection models. That could be like a `on_save(self, global_step, network)` or something like that.

```python
def on_save(self, global_step, args, network)
```

They also need to be loadable too but I think that can happen without another change.  

And we are needing to get access to the args a lot, might make sense to make it accessible in the network trainer directly under `self.args` then we wouldn't need to pass it around as much.

---

Here is showing the feature extraction from the model, but we'd have to call the `preprocess_transformer(flux2_model)` somewhere to add the feature extraction. 

```python
class FeatureExtractor:
    def __init__(self, n_features=2):
        self.features = [None] * n_features

    def register_feature_extractor(self, module: torch.nn.Module, n: int):
        def hook_fn(_module, _args, output):
            print("feature captured")
            self.features[n] = output.detach().clone()
            return output

        module.register_forward_hook(hook_fn)

class SelfFluxFlux2NetworkTrainer(Flux2NetworkTrainer):
    def __init__(self):
        super().__init__()

        self.extractor = FeatureExtractor()

    def preprocess_transformer(self, transformer: Flux2, args: argparse.Namespace):
        student_layer = args.student_feature_layer
        teacher_layer = args.student_feature_layer
        for i, (name, mod) in enumerate(flux.named_modules()):
            if i == student_layer:
                self.extractor.register_feature_extractor(mod, 0)

            if i == teacher_layer:
                self.extractor.register_feature_extractor(mod, 1)
```

I think if it can work more like this it will allow much more experimental interfaces on top of the tuner by third parties. Similar idea to how the network module itself can be done by third parties currently.

kohya-ss:
Hi @rockerBOO, thanks for the detailed proposal — your direction (Mixin + hooks, register_forward_hook, DiTOutput, etc.) is essentially the same extension mechanism I had in mind. The shape may differ in details once we discuss it fully, but designing that properly will take time, so I'd like to focus on a near-term landing strategy first.

I see two realistic paths and would like your input on which fits your situation better:

Option A — merge as-is (minimum work for you)
We merge PR #913 to main before I merge the dev refactor. That way you don't need to rebase onto the restructured trainer code — you just finalize PR #913 against current main. After it lands, I will reconcile the self-flow
changes into dev's new structure (i.e. move the additions from hv_train_network.py into training/trainer_base.py) on my side.

Under this option, it's fine either way whether self-flow logic stays inside the base trainer, or you apply some abstraction of your own (Mixin / subclass / etc.) before merge — please go with whichever is easiest for you. Just note that I will later redesign the extension mechanism and refactor self-flow onto it, so any abstraction introduced now is likely to be restructured. The user-facing CLI/behavior will be preserved as much as possible.

Option B — split + minimal seams (more work, but cleaner landing)

1. I merge dev to main first, so all subsequent work targets the new structure.
2. Independent features (plateau_logit_normal, gamma warmup, grad metrics, possibly the block_swap fix) are split out into small PRs and merged first.
3. I add a minimal seam to the base trainer — most likely process_batch() as the override point, plus a DiTOutput-style return — as no-op defaults.
4. You restructure the remaining self-flow code as a subclass / Mixin sitting on those seams (e.g. flux_2_train_network_self_flow.py).

The seam in step 3 would be intentionally minimal, not the final extension API, but designed so that later redesign can change internals without forcing you to rewrite the subclass.

Which would you prefer? If A, I'll keep the dev merge on hold until PR #913 lands. If B, I'd appreciate help with step 2 (the splits), and I can prepare step 3 in parallel.

That said, these are just the two options I came up with — if you see a better path, or have any concerns or thoughts about A or B, I'd really like to hear them before we decide. I want to pick the approach that works best for both of us.

rockerBOO:
Thank you for following up with your approaches. My initial approach was to do this extension approach, which I then had to augment the base network trainer code to properly implement this. The goal would be to create a working solution and then work back through it to get it accepted. I was not aware you were wanting to do the same refactoring I was envisioning. So we are on the same page here. 

With this, I have a working version with this PR. So going through making this work with the refactored approach before that is available would not be a good use of the time. I do not think anything in this PR should be blocking a future refactor. I would rather build on your refactor as you are developing it, so we can work through how well this would allow for the flexibility. And since I have done a bit of these experimental PR's I hope I can offer some perspective of how and what might need to be extendable. 

So to break it down, I would rather say this draft PR is static. It is usable in it's current state but not going to be merged as is. 

- We can pull some common functionality that I added here like grad metrics and possibly plateau_logit_normal into separate PRs. 
- DiTOutput was my initial idea but the larger concept is tuple outputs get really complicated after 1 or 2 outputs and  would work better in some struct/dict/dataclass format in general. This also allows us to freely add new variables to this without changing the datastructure

Only thing I would ask is to know when you are ready for experimenting/testing of the new refactoring. I have already separated the self-flow network trainer code so just would need to hook it in. Then we can iterate on what would be needed or augmented.

kohya-ss:
Thanks @rockerBOO, that direction works well for me — keeping PR #913 as a working reference and building Self-Flow on top of the refactor (rather than rebasing this PR) is a clean separation, and avoids wasted work on both sides.

On sequencing: I think it's better to finish the refactor + seams first, before you split out the shared pieces.
Otherwise the split PRs (grad metrics, plateau_logit_normal, etc.) would target code that's about to move, and we'd end up redoing most of the diff. Your "let me know when it's ready" comment is pointing at the same thing, I think.

Plan from my side:

1. The dev refactor is already fairly far along — hv_train_network.py → training/trainer_base.py separation is mostly done. What's left is adding the minimal seams (process_batch() as the override point, DiTOutput-style dataclass return from call_dit, plus probably a couple of hook points around sampling/saving/logging based on what you outlined).
2. I'll incorporate your extensibility suggestions from this PR while shaping those seams — the dataclass-over-tuple point especially, and the hook shape (on_sample, on_save, on_step_logs) you sketched out.
3. I'll prioritize getting this onto main so you're not blocked.

Once step 1 lands, you can:
- Split out the independent features (grad metrics, plateau_logit_normal, possibly the block_swap fix) as small PRs targeting the new structure.
- Hook the already-separated Self-Flow trainer onto the new seams and iterate.

Once the seams are merged into main (i.e. after your feedback round below), you can target main directly for both the split PRs and the Self-Flow extension — no need to track dev.

And — when I have the seam shapes drafted but before merging dev into main, I'd really like to get your feedback on them. You've been working through the actual extension surface in this PR, so your perspective on what needs to be extendable will be more concrete than what I'd come up with in isolation. I'll ping you at that point.

Does this work for you? If so, I'll get moving on the seam work and reach out when there's something concrete to look at.

rockerBOO:
Yes, this all works for me. 

My separation work is a non-factor as well. It was just helpful at describing what would be necessary for extending and minimal actual implementation beyond just moving methods over.

kohya-ss:
Hi @rockerBOO,

The dev refactor + extension seams are ready for your review: #943 (refactor/extension-seams → dev).

As we discussed, the seams are kept minimal and in-tree — plain subclass-override style, not a public extension protocol — so we keep the freedom to refactor later. The Phase 1/2 commits are all behaviour-preserving; the part I'd most like your eyes on is Phase 3.

The most useful place to look is `flux_2_train_network_self_flow.py`. It's a skeleton subclass of Flux2NetworkTrainer where every seam is overridden with shape-only stubs that reference the corresponding sections of your PR #913. The bodies are `NotImplementedError` / `# TODO` rather than actual logic — the file is not runnable yet. The intent is to make the seam surface concrete so you can tell whether each override actually has the data and lifecycle access you need to port your Self-Flow implementation onto it.

Two specific things I'd appreciate feedback on:

1. Coverage — is anything in your current Self-Flow implementation landing without a clean home in these seams? That's the most valuable signal, since now is the time to add or reshape a seam before it calcifies.

2. The one known open question, called out in the skeleton's call_dit docstring: per-token timesteps need support inside the FLUX.2 transformer model itself. I've treated that as a per-arch follow-up rather than a base-class seam concern, but please flag if you see it differently.
In the future, I think it would be good to be able to add some kind of extension point to the model file as well. However, I also think that can wait until a few such modifications occur.

Please don't feel any pressure on timing — I'm honestly the one who's usually slow to review, so this is on your schedule, not mine. I'll hold the dev → main merge until you've had a chance to look. Thanks for bearing with the wait this whole time.

Thanks!

PR #921, opened 2026-03-24
fix(cache_latents): progress bar total count

belambert:
This change adds a `total=` argument to a `tqdm` progress bar in the cache_latents code. This allows the progress bar to show the actual percent progress, rather than just counting up from 1.

I'm not sure if `get_total_image_count()` is always one-to-one equal to the number of latents computed (e.g. if there are duplicates). But for the local datasets I'm using, the total image count seemed to always be equal to the sum of the sizes of the batches.


kohya-ss:
Thank you for this! We will verify that `get_total_image_count` always returns the correct total sooner.

rockerBOO:
Added a test and seems to check out.

PR #943, opened 2026-05-06
Refactor: introduce extension seams on NetworkTrainer (for Self-Flow + future extensions)

kohya-ss:
## Background

Follow-up to the Self-Flow discussion in #913. As agreed there, instead of merging #913 as-is, the dev branch grows a small set of in-tree extension seams; Self-Flow then targets those seams as a subclass after dev → main.

This PR lands those seams on `dev`. **Not merging yet** — sharing first with @rockerBOO so he can flag any gaps in the seam shape before the design calcifies.

## Scope and design choice

- Mechanism: plain subclass-override, not a generic Extension protocol. Reasoning: in-tree only, fork-or-PR for third parties; YAGNI on multi-extension dispatch until we actually have two simultaneous extensions.
- All seams are no-op default. Vanilla training behaviour is preserved bit-for-bit on every architecture.
- Each seam carries an "Internal extension point — no API stability guarantees" comment so we keep the freedom to refactor.

## What's in this PR

**Phase 1 — extension seams on `NetworkTrainer`** (3 commits, all behaviour-preserving):

- `process_batch()` — replaceable inner training step (default = current vanilla flow matching).
- Lifecycle hooks: `on_train_start`, `on_post_optimizer_step`, `on_sample_images` (around-hook), `on_post_save`.
- Contributor hooks: `extra_trainable_params`, `extra_metadata`, `extra_step_logs`.

Naming convention: `on_post_*` = post-event additive hook, `extra_*` = returns dict/list to merge, `on_*` (no prefix) = lifecycle / around. CLI args and arg validation use the existing per-script `main()` and `handle_model_specific_args` patterns rather than new seams.

**Phase 2 — `call_dit` → `DiTOutput` dataclass** (1 commit):

- New `DiTOutput(pred, target, features=None, extra={})` dataclass.
- All 9 architecture `call_dit` overrides updated, plus `**kwargs` accepted for forward-compatible kwargs (e.g. `per_token_timesteps`, `hidden_features`).
- Two finetune entry scripts (`qwen_image_train.py`, `zimage_train.py`) updated at their direct `call_dit` call sites.

**Phase 3 — Self-Flow skeleton** (1 commit):

- New `flux_2_train_network_self_flow.py`: a `Flux2NetworkTrainer` subclass demonstrating how every seam gets used by Self-Flow.
- Each override is shape-only — bodies are `# TODO` + `raise NotImplementedError` referencing the corresponding lines in #913.
- The file is **not runnable yet**; it exists so @rockerBOO can judge whether the seams cover Self-Flow's actual needs before we commit to them on `main`.

## Test plan

- [x] All 9 per-arch trainers and both finetune entry scripts import cleanly
- [x] `DiTOutput` dataclass shape sanity-checked
- [x] Self-Flow skeleton parser composition + subclass MRO verified
- [x] Smoke-tested vanilla training paths on FLUX.2, Z-Image, FramePack (Phase 1)
- [x] Smoke-tested vanilla training paths on a subset of architectures (Phase 2)
- [x] Phase 4 (review responses ①–⑤): all 9 per-arch trainers and Self-Flow skeleton import cleanly after refactor
- [x] Phase 4 smoke-tested vanilla training paths on FLUX.2 and Z-Image (compute_loss extraction, sample_images hook relocation, force_sync_upload, on_transformer_loaded)

## What I'd like from review

@rockerBOO — primary asks:

1. **Does the seam set cover everything you need for Self-Flow?** Especially look at `flux_2_train_network_self_flow.py` and check whether each override actually has the data / lifecycle access it needs.
2. **One known open question, called out in `call_dit`'s docstring**: per-token timesteps need support inside the FLUX.2 transformer itself. That's a per-arch follow-up rather than a base-class seam concern, but flag if you see it differently.
3. Anything in your existing #913 implementation that doesn't have an obvious home in these seams — that's the most useful signal.

Once feedback is in and any gaps are addressed, this lands on `dev`, then `dev` → `main`, then you can target `main` directly with the split PRs (grad metrics, plateau_logit_normal, etc.) and Self-Flow on top of these seams.

### DIFF
--- a/src/musubi_tuner/training/trainer_base.py
+++ b/src/musubi_tuner/training/trainer_base.py
@@ -1077,6 +1077,52 @@ class NetworkTrainer:
 
     # endregion model specific
 
+    # region extension seams
+    # Internal extension points — no API stability guarantees.
+    # Subclasses live in this repo; if you fork, expect breakage on updates.
+
+    def process_batch(
+        self,
+        args: argparse.Namespace,
+        accelerator: Accelerator,
+        transformer,
+        network,
+        batch: dict[str, torch.Tensor],
+        latents: torch.Tensor,
+        noise: torch.Tensor,
+        noise_scheduler,
+        dit_dtype: torch.dtype,
+        network_dtype: torch.dtype,
+        vae,
+        global_step: int,
+    ) -> torch.Tensor:
+        """Compute scalar loss for one training batch (pre-backward).
+
+        Default implementation: vanilla flow matching.
+        Override to implement custom training logic (Self-Flow etc.).
+
+        ``latents`` is already scale-shifted; ``noise`` is already sampled.
+        """
+        noisy_model_input, timesteps = self.get_noisy_model_input_and_timesteps(
+            args, noise, latents, batch["timesteps"], noise_scheduler, accelerator.device, dit_dtype
+        )
+
+        weighting = compute_loss_weighting_for_sd3(
+            args.weighting_scheme, noise_scheduler, timesteps, accelerator.device, dit_dtype
+        )
+
+        model_pred, target = self.call_dit(
+            args, accelerator, transformer, latents, batch, noise, noisy_model_input, timesteps, network_dtype
+        )
+        loss = torch.nn.functional.mse_loss(model_pred.to(network_dtype), target, reduction="none")
+
+        if weighting is not None:
+            loss = loss * weighting
+
+        return loss.mean()
+
+    # endregion extension seams
+
     def train(self, args):
         if not self._validate_args_and_init(args):
             return
@@ -1780,27 +1826,20 @@ class NetworkTrainer:
                     # Sample noise that we'll add to the latents
                     noise = torch.randn_like(latents)
 
-                    # calculate model input and timesteps
-                    noisy_model_input, timesteps = self.get_noisy_model_input_and_timesteps(
-                        args, noise, latents, batch["timesteps"], noise_scheduler, accelerator.device, dit_dtype
+                    loss = self.process_batch(
+                        args,
+                        accelerator,
+                        transformer,
+                        network,
+                        batch,
+                        latents,
+                        noise,
+                        noise_scheduler,
+                        dit_dtype,
+                        network_dtype,
+                        vae,
+                        global_step,
                     )
-
-                    weighting = compute_loss_weighting_for_sd3(
-                        args.weighting_scheme, noise_scheduler, timesteps, accelerator.device, dit_dtype
-                    )
-
-                    model_pred, target = self.call_dit(
-                        args, accelerator, transformer, latents, batch, noise, noisy_model_input, timesteps, network_dtype
-                    )
-                    loss = torch.nn.functional.mse_loss(model_pred.to(network_dtype), target, reduction="none")
-
-                    if weighting is not None:
-                        loss = loss * weighting
-                    # loss = loss.mean([1, 2, 3])
-                    # # min snr gamma, scale v pred loss like noise pred, v pred like loss, debiased estimation etc.
-                    # loss = self.post_process_loss(loss, args, timesteps, noise_scheduler)
-
-                    loss = loss.mean()  # mean loss over all elements in batch
 
                     accelerator.backward(loss)
                     if accelerator.sync_gradients:
--- a/src/musubi_tuner/training/trainer_base.py
+++ b/src/musubi_tuner/training/trainer_base.py
@@ -800,7 +800,7 @@ class NetworkTrainer:
                 line += "#" * int(w / max_weighting * CONSOLE_WIDTH)
                 print(line)
 
-    def sample_images(self, accelerator: Accelerator, args, epoch, steps, vae, transformer, network, sample_parameters, dit_dtype):
+    def sample_images(self, accelerator: Accelerator, args, epoch, steps, vae, transformer, sample_parameters, dit_dtype):
         """architecture independent sample images"""
         if not should_sample_images(args, steps, epoch):
             return
@@ -810,8 +810,6 @@ class NetworkTrainer:
         if sample_parameters is None:
             logger.error(f"No prompt file / プロンプトファイルがありません: {args.sample_prompts}")
             return
-
-        self.on_before_sample_images(accelerator, args, epoch, steps, vae, transformer, network, sample_parameters, dit_dtype)
 
         distributed_state = PartialState()  # for multi gpu distributed inference. this is a singleton, so it's safe to use it here
 
@@ -860,8 +858,6 @@ class NetworkTrainer:
 
         transformer.switch_block_swap_for_training()
         clean_memory_on_device(accelerator.device)
-
-        self.on_after_sample_images(accelerator, args, epoch, steps, vae, transformer, network, sample_parameters, dit_dtype)
 
     def sample_image_inference(self, accelerator, args, transformer, dit_dtype, vae, save_dir, sample_parameter, epoch, steps):
         """architecture independent sample images"""
@@ -1210,7 +1206,6 @@ class NetworkTrainer:
         based on the generated samples.
         """
         pass
-
 
     def extra_trainable_params(
         self,
@@ -1914,9 +1909,19 @@ class NetworkTrainer:
                 os.remove(old_ckpt_file)
 
         def _do_sample(epoch_arg, steps_arg):
-            self.sample_images(
+            if not should_sample_images(args, steps_arg, epoch_arg):
+                return
+            self.on_before_sample_images(
                 accelerator, args, epoch_arg, steps_arg, vae, transformer, network, sample_parameters, dit_dtype
             )
+            try:
+                self.sample_images(
+                    accelerator, args, epoch_arg, steps_arg, vae, transformer, sample_parameters, dit_dtype
+                )
+            finally:
+                self.on_after_sample_images(
+                    accelerator, args, epoch_arg, steps_arg, vae, transformer, network, sample_parameters, dit_dtype
+                )
 
         # For --sample_at_first
         if should_sample_images(args, global_step, epoch=0):
--- a/src/musubi_tuner/training/trainer_base.py
+++ b/src/musubi_tuner/training/trainer_base.py
@@ -800,7 +800,7 @@ class NetworkTrainer:
                 line += "#" * int(w / max_weighting * CONSOLE_WIDTH)
                 print(line)
 
-    def sample_images(self, accelerator: Accelerator, args, epoch, steps, vae, transformer, sample_parameters, dit_dtype):
+    def sample_images(self, accelerator: Accelerator, args, epoch, steps, vae, transformer, network, sample_parameters, dit_dtype):
         """architecture independent sample images"""
         if not should_sample_images(args, steps, epoch):
             return
@@ -810,6 +810,8 @@ class NetworkTrainer:
         if sample_parameters is None:
             logger.error(f"No prompt file / プロンプトファイルがありません: {args.sample_prompts}")
             return
+
+        self.on_before_sample_images(accelerator, args, epoch, steps, vae, transformer, network, sample_parameters, dit_dtype)
 
         distributed_state = PartialState()  # for multi gpu distributed inference. this is a singleton, so it's safe to use it here
 
@@ -858,6 +860,8 @@ class NetworkTrainer:
 
         transformer.switch_block_swap_for_training()
         clean_memory_on_device(accelerator.device)
+
+        self.on_after_sample_images(accelerator, args, epoch, steps, vae, transformer, network, sample_parameters, dit_dtype)
 
     def sample_image_inference(self, accelerator, args, transformer, dit_dtype, vae, save_dir, sample_parameter, epoch, steps):
         """architecture independent sample images"""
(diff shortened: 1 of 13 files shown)


src/musubi_tuner/training/trainer_base.py
```diff
@@ -1731,16 +1898,28 @@ def save_model(ckpt_name: str, unwrapped_nw, steps, epoch_no, force_sync_upload=
             if args.huggingface_repo_id is not None:
                 huggingface_utils.upload(args, ckpt_file, "/" + ckpt_name, force_sync_upload=force_sync_upload)
 
+            self.on_post_save(args, accelerator, network, ckpt_name, save_dtype, metadata_to_save)
```
rockerBOO: if we need to do huggingface upload, then we'd also need the `force_sync_upload` variable

src/musubi_tuner/training/trainer_base.py
```diff
@@ -81,6 +82,22 @@
...
+
+    Internal extension point — no API stability guarantees. Fields beyond
+    ``pred`` and ``target`` are optional and used only by extension subclasses
+    (e.g. ``features`` for representation-alignment losses). Use ``extra`` as
+    a typed escape hatch for future fields without breaking signatures.
+    """
+
+    pred: torch.Tensor
+    target: torch.Tensor
+    features: Optional[torch.Tensor] = None
```
rockerBOO: Probably don't need features here. It is not part of the vanilla or normal flow. Adding additional values would then just go into the extra dict.

src/musubi_tuner/training/trainer_base.py
```diff
@@ -1072,11 +1089,157 @@ def call_dit(
...
+        )
+
+        weighting = compute_loss_weighting_for_sd3(
+            args.weighting_scheme, noise_scheduler, timesteps, accelerator.device, dit_dtype
+        )
+
+        output = self.call_dit(
+            args, accelerator, transformer, latents, batch, noise, noisy_model_input, timesteps, network_dtype
+        )
+        loss = torch.nn.functional.mse_loss(output.pred.to(network_dtype), output.target, reduction="none")
```
rockerBOO: This could become `loss, loss_metrics = self.compute_loss(output)`

then we'd return `return loss.mean(), loss_metrics` 

the loss metrics would allow us to add the metrics to the logs.

src/musubi_tuner/training/trainer_base.py
```diff
@@ -1731,16 +1898,28 @@ def save_model(ckpt_name: str, unwrapped_nw, steps, epoch_no, force_sync_upload=
...
 
+            self.on_post_save(args, accelerator, network, ckpt_name, save_dtype, metadata_to_save)
+
         def remove_model(old_ckpt_name):
             old_ckpt_file = os.path.join(args.output_dir, old_ckpt_name)
             if os.path.exists(old_ckpt_file):
                 accelerator.print(f"removing old checkpoint: {old_ckpt_file}")
                 os.remove(old_ckpt_file)
 
+        def _do_sample(epoch_arg, steps_arg):
```
rockerBOO, line 1962: if on_sample_images is to extended then we need to pass all the same parameters to on_sample_images. in this example we'd also need epoch_arg, steps_arg, vae, transformer, sample_parameters, dit_dtype

I think part of the problem is a bit of misdirection here and some convoluted flow for sampling images. also sampling might not just be for images for other types of models?

The thing I'm trying to do in self-flow is change a value sent to the sampling, the network. We are also doing "network" and then that would turn into "transformer" later but we do not have control over what that is being sent to the sample_images directly. 

I am still thinking through how it might be better, but just feel this whole flow is not great. 

---

As I was writing this, I went and reviewed the sample images flow. The key change was in `sample_images` we'd have a before `on_before_sample_images` and after `on_before_sample_images` instead of `on_sample_images` to do the necessary hooks. 

There is more detail in my PR #944 to show more of those details about this abstraction.

rockerBOO (commented):
I haven't tested the code in practice yet but this is my feedback so far.

kohya-ss:
Thank you for your review! I will work on them ASAP.

rockerBOO:
One last bit I remembered was needing to be able to "hook" into the models after they are loaded.

```
def on_model_load(self, transformer)
```

but would need a load for each model type (text encoder, VAE). they possibly could use the same `on_model_load` though, with maybe a type hint (dit, text_encoder, vae). This is where I'd be hooking in my feature extraction register-forward_hook into the flux model.

src/musubi_tuner/training/trainer_base.py
kohya-ss: That's right. I will add `force_sync_upload` to the hook signature so subclasses can trigger sync uploads when needed.

src/musubi_tuner/training/trainer_base.py
kohya-ss: I agreed. `features` is extension-only and shouldn't be in the vanilla return type. I'll remove it, so extension subclasses can use the `extra` dict for this purpose.

src/musubi_tuner/training/trainer_base.py
kohya-ss: Adopting this — the metrics surface is genuinely useful. Plan:

- New `compute_loss(args, output, timesteps, noise_scheduler, network_dtype) -> tuple[Tensor, dict[str, float]]` that owns the full loss computation including weighting and reduction. I'd rather have it cover the whole thing than just the per-element MSE, since weighting is tightly coupled to the vanilla flow-matching loss form and Self-Flow-style losses will likely want to swap the entire formulation.
- `process_batch` returns `(loss, loss_metrics)`; the train loop does `logs.update(loss_metrics)` alongside the existing `extra_step_logs` merge (keeping them as separate concerns — loss decomposition vs. arbitrary subclass logs).
- Default `loss_metrics` stays minimal; subclasses populate as needed.
- Vanilla behavior remains bit-identical.

src/musubi_tuner/training/trainer_base.py
kohya-ss: Fully agree the around-hook is awkward. Adopting `on_before_sample_images` / `on_after_sample_images` with `try/finally` so the after-hook always runs, and passing the loose args (args, accelerator, epoch, steps, network, transformer, vae, dit_dtype, sample_parameters) directly to the hooks.

Re: #944 — thanks for the concrete prototype, that made the discussion much easier. I've replied there in more detail, but the short version: I'd like to take the hook split with loose positional args and hold off on `SamplingContext` / `SamplePrompt` for now to keep #943's scope minimal. Happy to revisit the dataclass refactor as a follow-up once more concrete extension cases drive the shape.

kohya-ss:
Thank you! That's nice. I will add this.

Scoping to the transformer for now:

`on_transformer_loaded(self, args, accelerator, transformer)` called right after `self.load_transformer(...)` in the trainer flow — defaults to no-op. That should cover the FLUX `register_forward_hook` case you mentioned.

Holding off on text-encoder / VAE variants for now: text encoders are loaded per-architecture in pretty different ways (often via `transformers` library directly, and only at sample-prompt prep time rather than in the main trainer flow), and there's no concrete VAE-hook need yet. Easy to add later if a real use case shows up — keeping the surface minimal until then.

kohya-ss:
Thank you for your feedback on #945!

I believe I've updated all the review points. I would appreciate it if you could review this again.

rockerBOO:
I think at this point all my initial feedback has been addressed. For self-flow I do not see any glaring issues. I wanted to do a test implementation for self-flow using this but I don't think it is blocking. For any future hooks, I think we are on the same page of how that would work so I can  propose those as I run into issues.

src/musubi_tuner/training/trainer_base.py
```diff
@@ -1072,11 +1089,157 @@ def call_dit(
...
+        Default implementation: vanilla flow matching.
+        Override to implement custom training logic (Self-Flow etc.).
+
+        ``latents`` is already scale-shifted; ``noise`` is already sampled.
+        """
+        noisy_model_input, timesteps = self.get_noisy_model_input_and_timesteps(
+            args, noise, latents, batch["timesteps"], noise_scheduler, accelerator.device, dit_dtype
+        )
+
+        weighting = compute_loss_weighting_for_sd3(
```
rockerBOO: this weighting, is it for all flow models? maybe makes sense as a separate option or inside `compute_loss`?

src/musubi_tuner/training/trainer_base.py
```diff
@@ -1072,11 +1089,157 @@ def call_dit(
...
+
+        Use this for initializing extension state that depends on prepared models
+        (EMA copies, decay schedulers, register_forward_hook on the transformer, etc.).
+        """
+
+    def on_post_optimizer_step(
+        self,
+        args: argparse.Namespace,
+        accelerator: Accelerator,
+        network,
```
rockerBOO, line 1203: We are applying the network here, but why not the other models? At least transformer here makes sense if we are training that (for non-network training)?

src/musubi_tuner/training/trainer_base.py
```diff
@@ -1072,11 +1089,157 @@ def call_dit(
...
+        ``sync_gradients`` mirrors ``accelerator.sync_gradients`` and is True only
+        on steps where an actual optimizer update occurred (gradient accumulation aware).
+        Use for EMA updates or any post-step bookkeeping.
+        """
+
+    def on_sample_images(
+        self,
+        args: argparse.Namespace,
+        accelerator: Accelerator,
+        network,
```
rockerBOO: Same idea as the other review about network, and not including the other models.

src/musubi_tuner/training/trainer_base.py
```diff
@@ -1072,11 +1089,157 @@ def call_dit(
...
+        temporarily swap student weights for EMA (teacher) weights before sampling
+        and restore them afterwards.
+        """
+        sample_fn()
+
+    def on_post_save(
+        self,
+        args: argparse.Namespace,
+        accelerator: Accelerator,
+        network,
```
rockerBOO, line 1222: Same as the previous about the models.

rockerBOO (commented):
Some smaller function parameter things to consider. And weighting inside process_batch but probably beyond the extension seams in this PR.

rockerBOO:
Also ran through my wavelet loss through the new hooks and formulation if you want to see <https://github.com/kohya-ss/musubi-tuner/compare/refactor/extension-seams...rockerBOO:musubi-tuner:example-wavelet-loss> . Only real issue I found is needing the weighting inside the `compute_loss` so im re-calculating it inside. 

Edit: Additionally thought about what happens when we have multiple levels of extensions. Like if i wanted self-flow and wavelet loss together, it really falls apart. Might be worth exploring what options are available for this, which might be where we can "register" hooks and then it will run all the hooks when it is typically doing 1. 

There may be other options. Let me know what you think about that and if you have any ideas. I can look a little further as well.

src/musubi_tuner/training/trainer_base.py
kohya-ss: I agree it belongs inside `compute_loss.` That's already done in the latest revision (the `compute_loss` extraction commit), so per-element weighting now lives next to the MSE.

src/musubi_tuner/training/trainer_base.py
kohya-ss: It now takes transformer as well.

src/musubi_tuner/training/trainer_base.py
kohya-ss: Good catch, the asymmetry there wasn't intentional. I've added transformer to both signatures (and updated the Self-Flow override accordingly) so non-network / full fine-tuning subclasses, or anything wanting to do EMA/companion-save against transformer weights, can reach it without stashing a reference in `on_train_start`.

kohya-ss:
Recomputing `compute_loss_weighting_for_sd3` inside the overridden `compute_loss` is, I think, the right shape here rather than something I should help with from the base class. The SD3 weighting scheme isn't universally applicable to every DiT architecture as you wrote, that's actually the reason it lives in `compute_loss` rather than `process_batch` now.

> Edit: Additionally thought about what happens when we have multiple levels of extensions. Like if i wanted self-flow and wavelet loss together, it really falls apart. Might be worth exploring what options are available for this, which might be where we can "register" hooks and then it will run all the hooks when it is typically doing 1.
> 
> There may be other options. Let me know what you think about that and if you have any ideas. I can look a little further as well.

This one I'd like to leave out of scope, with a clear "won't do" rather than "not yet." A register-hook style fan-out across multiple independent extensions can't really guarantee correct behaviour when the user combines arbitrary pieces — ordering, mutual data dependencies, and conflicting state assumptions (e.g. an EMA-swapping hook vs. a sampling-time auxiliary loss) all become silent footguns. The deliberate design choice here is: if you want to combine two extensions, do it by composition.

Concretely, the pattern I'd recommend is to keep each extension's actual logic in plain, standalone functions (or a small helper class that doesn't own the training loop), and have the trainer subclass be a thin shell that calls them from the appropriate hooks. Then a "Self-Flow + wavelet-loss" trainer is just a third subclass that delegates to both helpers in the order it explicitly chose — not a multiple-inheritance hop that hopes MRO does the right thing. That keeps the combination something a human has thought about and tested, rather than something the framework guesses at runtime; your own observation that Self-Flow + wavelet-loss "really falls apart" through naive composition is exactly the failure mode I'd rather surface at write-time than paper over with a registry.

Happy to revisit if a concrete combined use case shows up where the delegation route is genuinely painful, but I'd want to see that case before adding the machinery.

rockerBOO:
Thank you for the clarity if you were interested in that approach. I think those issues make things much more complicated.

I did a little exploring of the extension orchestration and came up with an example. There are core parts that need to be reasoned about like mixing process_batch between 2 different runs extensions which handle this with a forward pass in each. So mixing them will require  knowing how they both will handle this. 

https://github.com/kohya-ss/musubi-tuner/compare/0e6c554f396f46dddcd07ffdd9241efddfccbcdb...rockerBOO:musubi-tuner:feature/trainer-orchestrator

This example is just for exploration, something that could be implemented externally or for experimentation.

kohya-ss:
Thanks for taking the time to explore this — the four-way classification is well thought out, and in particular making `process_batch` / `compute_loss` raise on conflict (rather than silently merging) lines up with the "surface it at write-time" stance from earlier. Nice touch.

I think your own framing of it — "exploration, something that could be implemented externally or for experimentation" — is exactly right, and I'd like to keep it out of the core repo on the same grounds I mentioned before: the void hooks still fan out in registration order, so order-sensitive combinations (e.g. an EMA-swap hook vs. a post-step auxiliary loss) remain implicit, and I'd rather a combined trainer be a small explicit subclass that the author has actually thought about.

PR #944, opened 2026-05-07
Refactor: typed SamplingContext + SamplePrompt for sample_images flow

rockerBOO:
## Summary

- Introduces `SamplingContext` and `SamplePrompt` dataclasses (`training/sampling.py`), replacing 8–9 loose positional args in `sample_images` / `sample_image_inference`
- Replaces the lambda around-hook `on_sample_images` with simple lifecycle hooks `on_before_sample_images` / `on_after_sample_images`, consistent with existing hook style (`on_train_start`, `on_post_optimizer_step`)
- Migrates all 10 architectures' `process_sample_prompts` to return `list[SamplePrompt]` instead of `list[dict]`; `do_inference` receives `SamplePrompt` directly (attribute access throughout)
- Self-flow skeleton updated: EMA weight swap now uses `_saved_student_state` with the before/after hook pattern, with `ctx.network` and `ctx.accelerator` explicit in the hook signature

## Design notes

`SamplingContext` carries all shared sampling state (accelerator, transformer, network, vae, etc.). `SamplePrompt` is flat with all known architecture-specific fields as optional — subclasses can extend it later. `discrete_flow_shift` is `float | None` to accommodate FLUX.2's sentinel `None`.

Known follow-up: `SamplePrompt` has duplicate `vl_embed`/`negative_vl_embed` field declarations (Qwen-Image and HunyuanVideo1.5 both use the same name — works at runtime, worth disambiguating in a future pass). `do_inference` still receives redundant unpacked positional args alongside the `SamplePrompt` object — future refactor opportunity.

## Goals

The rest of this PR was written by Claude but to give some of my perspective which I went through before implementing this.

- `SamplePrompt` was made to unify the sample prompt data structure. This got a little longer than intended where I would want it to be architecture specific that extends SamplePrompt. 
- `SamplingContext` was made to put all the variables together instead of a very long argument string to make it have a known data structure and easier to pass around and extend this behavior. 

If the rest of this PR is ignored, then the only thing necessary is the `on_before_sample_images` and `on_after_sample_images` inside the `sample_images` would be the desired approach. 

## Test plan

- [x] All 7 unit tests pass (`uv run pytest tests/`)
- [ ] Verify sample image generation works end-to-end for FLUX.2, WAN, HunyuanVideo
- [ ] Verify self-flow EMA swap path (`--self_flow`) generates samples with teacher weights

### DIFF
--- a/docs/plans/2026-05-07-sample-images-refactor-plan.md
+++ b/docs/plans/2026-05-07-sample-images-refactor-plan.md
@@ -1,0 +1,708 @@
+# Sample Images Refactor Implementation Plan
+
+> **For Claude:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task.
+
+**Goal:** Replace 8–9 loose positional args in `sample_images`/`sample_image_inference` with typed `SamplingContext` + `SamplePrompt` dataclasses, and replace the lambda around-hook with simple before/after lifecycle hooks.
+
+**Architecture:** New `sampling.py` introduces both dataclasses. `trainer_base.py` adopts them and swaps the around-hook for two simple hooks. Architecture-specific `process_sample_prompts` methods return `list[SamplePrompt]` instead of `list[dict]`. `do_inference` is untouched throughout.
+
+**Tech Stack:** Python 3.10+, dataclasses, pytest (`uv run --extra cu130 pytest`)
+
+**Design doc:** `docs/plans/2026-05-07-sample-images-refactor-design.md`
+
+---
+
+## Task 1: Add `sampling.py` with `SamplePrompt` and `SamplingContext`
+
+**Files:**
+- Create: `src/musubi_tuner/training/sampling.py`
+- Create: `tests/training/test_sampling.py`
+
+**Step 1: Write the failing test**
+
+```python
+# tests/training/test_sampling.py
+import torch
+import pytest
+from unittest.mock import MagicMock
+
+
+def test_sample_prompt_defaults():
+    from musubi_tuner.training.sampling import SamplePrompt
+    p = SamplePrompt(prompt="test", width=512, height=512)
+    assert p.prompt == "test"
+    assert p.width == 512
+    assert p.height == 512
+    assert p.sample_steps == 20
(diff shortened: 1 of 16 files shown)


kohya-ss:
Thanks for putting this together, having a concrete prototype made the discussion in #943 much easier to ground. The before/after hook split in particular is a clear win over the around-hook I had, and I'll be adopting it in #943.

On `SamplingContext` and `SamplePrompt`, I'd like to hold off for now and keep #943 scoped to the minimum seam set. A few reasons:

- **`SamplePrompt`** ripples through all 10 architectures' `process_sample_prompts` and `do_inference`, and the flat-with-Optionals shape is already showing tension (the duplicated `vl_embed` / `negative_vl_embed` you flagged is a symptom. Qwen-Image and HunyuanVideo 1.5 wanting the same field name suggests architecture-specific subclasses would  be the right shape, but that's a larger design conversation than #943 should host).
- **`SamplingContext`** is more contained (trainer_base only), but introducing it alongside the hook split makes #943's scope harder to reason about. I'd rather land the hooks with loose positional args first, and come back to context-objects when there's a concrete second extension pushing on the shape.
- General preference for adding seams as use cases concretize, rather than ahead of them.

So the plan is: take `on_before_sample_images` / `on_after_sample_images` (with loose args) into #943, leave `SamplePrompt` / `SamplingContext` aside for now. Definitely not closing the door on the dataclass refactor — happy to revisit as a follow-up PR once we have more signal on what shape fits.

One small thing: `docs/plans/2026-05-07-sample-images-refactor-plan.md` looks like it might have been committed by accident — was that intentional?

Thanks again for the prototype and the careful thinking — really helpful even where I'm not taking the changes wholesale. Whether you'd like to keep this open as a reference, repurpose it for the hook split portion, or close it is entirely up to you.

rockerBOO:
Thanks for reviewing this PR! I agree that all the extra changes would not be appropriate at this time. I still feel having the data structure would make things easier so would like to revisit this later. Going to make this a draft and not intent on taking this further until later where we can revisit this. 

I have made the changes in #945 that should be the minimal implementation using our current format.

PR #945, opened 2026-05-08
Add on_before_sample_images and on_after_sample_images

rockerBOO:
As per feedback in #944 and building on #943 adds the simple abstraction for `on_before_sample_images` and `on_after_sample_images`

The position in `sample_images` may need to be tuned to where most appropriate but I think this works.

### DIFF
--- a/src/musubi_tuner/flux_2_train_network_self_flow.py
+++ b/src/musubi_tuner/flux_2_train_network_self_flow.py
@@ -225,25 +225,21 @@ class Flux2SelfFlowNetworkTrainer(Flux2NetworkTrainer):
         # TODO: in-place lerp_ across floating-point params; copy_ otherwise.
         raise NotImplementedError("Self-Flow EMA update — see PR #913 _update_ema_weights")
 
-    def on_sample_images(
-        self,
-        args: argparse.Namespace,
-        accelerator: Accelerator,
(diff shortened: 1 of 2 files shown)


kohya-ss:
Merged — thanks for the quick turnaround on this!

Just a heads up: I pushed a small follow-up (a6f8e86) that:

1. Wraps the hooks with `try/finally` in `_do_sample` so `on_after_sample_images`runs even if sampling raises. This matters for Self-Flow's EMA swap — without it, a sampling failure leaves the network in EMA state with `_saved_student_state` still populated, which would corrupt the next swap cycle.

2. Moves the hook calls out of `sample_images` itself and into `_do_sample`, restoring `sample_images`'s original signature (no `network` parameter). The fine-tuning entry scripts `qwen_image_train.py` / `zimage_train.py` call `sample_images` directly without a `network`, so the added parameter would have broken them. `_do_sample` already has `network` in its closure scope, so calling the hooks there keeps `sample_images` clean.

The hook signatures themselves are unchanged, so the Self-Flow skeleton override still works as-is. Both refactors are small enough that I figured it was simpler to ship them as a follow-up commit than to round-trip through review again.

Thanks again — the before/after split is much nicer than the around-hook I had originally.

PR #946, opened 2026-05-11
Add dual-GPU model-parallel path for FLUX.2 + Wan 2.2 + HunyuanVideo LoRA training

genno-whittlery:
## Summary

Adds an env-var-gated **dual-GPU model-parallel** path for **FLUX.2, Wan 2.2, and HunyuanVideo** LoRA training. Splits the transformer blocks at the midpoint across `cuda:0`/`cuda:1`, letting users train on pairs of 24+ GB consumer GPUs (2× RTX 3090 / 4090 / 5090) where the single-GPU path WDDM-thrashes or doesn't fit.

Off by default — no behavior change unless `FLUX2_DUAL_GPU=true` (FLUX.2), `WAN_DUAL_GPU=true` (Wan 2.2), or `HV_DUAL_GPU=true` (HunyuanVideo) is set.

> Originally filed as FLUX.2-only. Wan 2.2 and HunyuanVideo support are folded into the same PR because all three ports edit the same shared infrastructure (`networks/lora.py`, `hv_train_network.py`) — keeping them as separate PRs would guarantee merge conflicts. The dual-GPU machinery is genuinely shared; this PR is the per-trainer unit of "does musubi-tuner support dual-GPU model-parallel".

### FLUX.2
- `src/musubi_tuner/flux_2/flux2_dual_gpu.py` (new): ~200 LOC helper. Distributes `Flux2Transformer2DModel.single_transformer_blocks`, registers per-block forward pre-hooks (every cuda:1 block, not just the boundary, because FLUX.2's forward passes loop-level constants `temb_mod_params` / `image_rotary_emb` to each block), wraps the split forward in `torch.amp.autocast(cuda, bf16)`, tuple-aware boundary transfer for `single_block_mod`.
- `src/musubi_tuner/flux_2/flux2_models.py`: integration hook.

### Wan 2.2
- `src/musubi_tuner/wan/wan_dual_gpu.py` (new): ~290 LOC helper. Distributes the vendored `WanModel` — `patch_embedding` / `text_embedding` / `time_embedding` / `time_projection` / `img_emb` + `blocks[:split_at]` + `head` on cuda:0, `blocks[split_at:]` on cuda:1. `freqs` and the `freqs_fhw` cache are plain tensor attributes (`WanModel` deliberately avoids `register_buffer` to keep dtype stable across `.to()`), so they're moved to cuda:0 explicitly. Split-aware forward replaces `WanModel.forward`: `WanModel` builds one `kwargs` dict and reuses it for every `block(x, **kwargs)` call, so a single PCIe boundary crossing at `blocks[split_at]` (bridge `x` + the shared `kwargs` dict to cuda:1) feeds all downstream blocks — no per-block hooks needed. `enable_wan_dual_gpu` is idempotent (safe after an A14B high/low expert swap); guards against `<2` CUDA devices and `--blocks_to_swap` (mutually exclusive with the split).
- `src/musubi_tuner/wan/modules/model.py`: `move_to_device_except_swap_blocks()` early-returns through `enable_wan_dual_gpu` when `WAN_DUAL_GPU=true`.
- `src/musubi_tuner/wan_train_network.py`: `load_transformer()` fail-fast `<2`-GPU guard before the slow checkpoint load.

### HunyuanVideo
- `src/musubi_tuner/hunyuan_video_dual_gpu.py` (new): ~327 LOC helper. `hv_train_network.py` is musubi-tuner's native trainer (the `NetworkTrainer` base the FLUX.2 / Wan trainers subclass), and HunyuanVideo's `HYVideoDiffusionTransformer` has a double-stream + single-stream structure closer to FLUX.2 than Wan. `enable_hv_dual_gpu` distributes input projections / embedders + all 20 `double_blocks` + `single_blocks[:split_at]` on cuda:0, `single_blocks[split_at:]` + `final_layer` on cuda:1 (default `split_at = len(single_blocks)//2 = 20`). Single-boundary forward replacement: the `double_blocks` loop runs entirely on cuda:0, then after `x = torch.cat((img, txt), 1)` the `single_blocks` loop inserts one cuda:1 boundary — `_bridge_block_args` recurses the per-block positional arg lists (handles the `freqs_cis` tuple). Fewer gotchas than Wan: RoPE is passed as forward args (not a plain model attribute), and the two text encoders are pre-cached + `del`'d before the training loop, so no `text_encoder_to` handling is needed. Guards against `<2` CUDA devices and `--blocks_to_swap`.
- `src/musubi_tuner/hunyuan_model/models.py`: `move_to_device_except_swap_blocks()` early-returns through `enable_hv_dual_gpu` when `HV_DUAL_GPU=true`.

### Shared infrastructure (touched by both)
- `src/musubi_tuner/hv_train_network.py`: three integration-point edits, each gated on `FLUX2_DUAL_GPU` / `WAN_DUAL_GPU` / `HV_DUAL_GPU` — DDP bypass (skip `InitProcessGroupKwargs`), CPU loading device (so the large bf16 transformer doesn't pre-allocate on cuda:0 before the split), `device_placement=[False]` branch in `accelerator.prepare`.
- `src/musubi_tuner/networks/lora.py`: `LoRAModule.apply_to` snapshots the wrapped layer's device; `LoRAModule.forward` inline-pins LoRA params to the wrapped layer's device, gated on any of the three env vars. The bound-method indirection in `apply_to` (`self.org_module.forward = self.forward`) makes instance-level monkey-patching unreliable, so the pin runs inline. The inline pin also lazily re-places LoRA params after a Wan 2.2 A14B high/low expert state-dict swap. A flux2-gated eager `route_loras_to_wrapped_devices` pass remains for the FLUX.2 path.

## Validation

Both ports validated end-to-end on 2× RTX 5090:

**FLUX.2** — sumi v8 (32 imgs, 512²), `--fp8_base --fp8_scaled --gradient_checkpointing`, rank 32:
- 10/10 steps, loss 0.526 → 0.545 monotonic, **2.22 s/it sustained**
- Distribution: 20.7 GB cuda:0 / 12.6 GB cuda:1 after fp8 weight-only quant

**Wan 2.2** — T2V-A14B, `--task t2v-A14B --fp8_base --network_module networks.lora_wan`, rank 16:
- single-expert: `distributed 40 blocks across cuda:0/cuda:1 (20/20)`, 10/10 steps, ~1.0–1.6 it/s, loss stable
- dual-expert (`--dit` + `--dit_high_noise` + `--timestep_boundary`): i2v-A14B, 20 steps, 3 distribution events logged (1 initial + 2 swap-driven re-distributes), checkpoint saved

**HunyuanVideo** — `--fp8_base --network_module networks.lora`, rank 16:
- `enable_hv_dual_gpu` log: `distributed 20 double + 40 single blocks (cuda:0: 20 double + 20 single; cuda:1: 20 single + final_layer)`
- 20/20 steps, **~1.30 it/s**, avr_loss 0.155 → 0.217, checkpoint saved

## Context

This dual-GPU distribution shape has been validated across multiple FLUX.2 LoRA trainers (ai-toolkit, musubi-tuner, OneTrainer, DiffSynth-Studio) and Wan 2.2 + LTX-2 + Qwen-Image (ai-toolkit) and Wan 2.2 + HunyuanVideo (musubi-tuner). Cross-trainer write-up: https://github.com/genno-whittlery/flux2-dual-gpu-lora — musubi-tuner porting walkthrough: https://github.com/genno-whittlery/flux2-dual-gpu-lora/blob/main/docs/porting-musubi-tuner.md

## Test plan
- [x] FLUX.2 sumi v8 10-step smoke on 2× RTX 5090, fp8 base+scaled, gradient checkpointing, rank 32 — loss monotonic, checkpoint saves
- [x] Wan 2.2 T2V-A14B 10-step smoke on 2× RTX 5090, fp8 base, rank 16 — 40 blocks split 20/20, loss stable, checkpoint saves
- [x] HunyuanVideo 20-step smoke on 2× RTX 5090, fp8 base, rank 16 — 20 double + 40 single blocks distributed, loss stable, checkpoint saves
- [x] single-GPU path unchanged (neither env var set → same `hv_train_network.py` code path)
- [ ] cross-GPU-vendor validation (2× RTX 3090, 2× RTX 4090) — pending
- [x] Wan 2.2 A14B dual-expert path (`--dit` + `--dit_high_noise` + `--timestep_boundary`) — validated; the swap re-distributes via an idempotent `enable_wan_dual_gpu` + dtype-fixup in `swap_high_low_weights`

### DIFF
--- a/src/musubi_tuner/flux_2/flux2_dual_gpu.py
+++ b/src/musubi_tuner/flux_2/flux2_dual_gpu.py
@@ -1,0 +1,263 @@
+"""Dual-GPU model-parallel path for FLUX.2 LoRA training in musubi-tuner.
+
+Activates when ``FLUX2_DUAL_GPU=true``. Distributes the FLUX.2 transformer
+across two CUDA devices with a single PCIe boundary mid-``single_blocks``,
+and routes per-layer LoRA modules to the device of the transformer layer
+they wrap. Default behavior is unchanged when the env var is unset.
+
+Companion to the validated ai-toolkit patch
+(https://github.com/genno-whittlery/flux2-dual-gpu-lora). Musubi-tuner's
+architecture is structurally cleaner for porting:
+
+- Text encoders (Mistral 3 / Qwen-3) are pre-cached to disk via the
+  separate ``flux_2_cache_text_encoder_outputs.py`` script, so the inner
+  training loop never touches them. No TE-on-CPU runtime patch needed.
+- ``LoRAModule.apply_to`` is a 3-line monkey-patch over ``forward``; a
+  one-line snapshot of the wrapped layer's device plus a lazy forward
+  shim is sufficient for per-LoRA device routing.
+- ``Flux2.move_to_device_except_swap_blocks`` is the single canonical
+  place to intercept transformer-wide device placement.
+
+Env vars:
+    FLUX2_DUAL_GPU=true                    enable the dual-GPU path
+    FLUX2_DUAL_GPU_SPLIT_AT=24             override single_blocks split
+                                           index (default: n_single // 2)
+"""
+from __future__ import annotations
+
+import os
+from typing import TYPE_CHECKING, Callable, Optional
+
+import torch
+import torch.nn as nn
+
+if TYPE_CHECKING:
+    from musubi_tuner.flux_2.flux2_models import Flux2
+    from musubi_tuner.networks.lora import LoRANetwork, LoRAModule
+
+
+def is_dual_gpu_enabled() -> bool:
+    """True iff ``FLUX2_DUAL_GPU=true`` in the environment."""
+    return os.getenv("FLUX2_DUAL_GPU", "false").lower() == "true"
+
+
+def get_split_at(num_single_blocks: int) -> int:
+    """Single-blocks split index. Override via ``FLUX2_DUAL_GPU_SPLIT_AT``."""
+    override = os.getenv("FLUX2_DUAL_GPU_SPLIT_AT")
+    if override is not None:
+        return int(override)
+    return num_single_blocks // 2
+
+
+def assert_no_block_swap_in_dual_gpu(blocks_to_swap: Optional[int]) -> None:
+    """Block-swap and dual-GPU are mutually exclusive.
+
+    Block-swap pages whole blocks RAM↔single-GPU; dual-GPU splits blocks
+    across two GPUs statically. The two strategies make different
+    assumptions about block residence and don't compose.
+    """
+    if not is_dual_gpu_enabled():
+        return
+    if blocks_to_swap and blocks_to_swap > 0:
+        raise RuntimeError(
+            "FLUX2_DUAL_GPU=true is incompatible with --blocks_to_swap. "
+            "The dual-GPU split already distributes blocks across two GPUs; "
+            "block-swap pages blocks to system RAM, which fights the split. "
+            "Disable --blocks_to_swap or unset FLUX2_DUAL_GPU."
+        )
+
+
+def distribute_flux2_transformer(model: "Flux2", dtype: torch.dtype) -> None:
+    """Distribute the FLUX.2 transformer across cuda:0 and cuda:1.
+
+    Layout:
+        cuda:0  — img_in, txt_in, time_in, pe_embedder, guidance_in,
+                  *_modulation*, all double_blocks, single_blocks[0:split_at]
+        cuda:1  — single_blocks[split_at:], final_layer
+
+    The forward path is updated by :func:`install_split_forward` to insert
+    one ``.to(cuda:1)`` boundary at the split point and ``.to(cuda:0)``
+    on the way back to ``final_layer`` (we keep final_layer on cuda:0 so
+    the returned tensor matches ``vec.device`` for the caller).
+    """
+    if torch.cuda.device_count() < 2:
+        raise RuntimeError(
+            f"FLUX2_DUAL_GPU=true requires ≥2 CUDA devices, found "
+            f"{torch.cuda.device_count()}."
+        )
+
+    split_at = get_split_at(model.num_single_blocks)
+    if not 0 < split_at < model.num_single_blocks:
+        raise RuntimeError(
+            f"FLUX2_DUAL_GPU_SPLIT_AT={split_at} out of range "
+            f"(model has {model.num_single_blocks} single blocks)."
+        )
+
+    cuda0 = torch.device("cuda:0")
+    cuda1 = torch.device("cuda:1")
+
+    # Pre-blocks scaffolding + all double_blocks + first half single_blocks
+    # land on cuda:0. Final layer also lives on cuda:0 (see forward).
+    model.img_in.to(cuda0, dtype=dtype)
+    model.txt_in.to(cuda0, dtype=dtype)
+    model.time_in.to(cuda0, dtype=dtype)
+    model.pe_embedder.to(cuda0)
+    if model.use_guidance_embed and not isinstance(model.guidance_in, nn.Identity):
+        model.guidance_in.to(cuda0, dtype=dtype)
+    model.double_stream_modulation_img.to(cuda0, dtype=dtype)
+    model.double_stream_modulation_txt.to(cuda0, dtype=dtype)
+    model.single_stream_modulation.to(cuda0, dtype=dtype)
+    for block in model.double_blocks:
+        block.to(cuda0, dtype=dtype)
+    for block in model.single_blocks[:split_at]:
+        block.to(cuda0, dtype=dtype)
+    for block in model.single_blocks[split_at:]:
+        block.to(cuda1, dtype=dtype)
+    # final_layer stays on cuda:0 — the forward moves img back before
+    # calling it, and final_layer's output matches vec.device (cuda:0).
+    model.final_layer.to(cuda0, dtype=dtype)
+
+    # Stash the split index for the patched forward.
+    model._flux2_dual_gpu_split_at = split_at  # type: ignore[attr-defined]
+
+
+def install_split_forward(model: "Flux2") -> None:
+    """Replace ``Flux2.forward`` with a split-aware version.
+
+    Inserts one ``.to(cuda:1)`` boundary at ``single_blocks[split_at]``
+    and ``.to(cuda:0)`` after the loop so ``final_layer`` (and the
+    returned tensor) live on cuda:0 — matching ``vec.device`` for the
+    rest of the training step.
+
+    Cross-PCIe traffic per forward: one activation tensor
+    (~18 MB at 512², bf16) one way, plus the same on the way back
+    during backward. At PCIe Gen5 x16 (~50 GB/s practical), this is
+    ~1 ms — well below the per-step compute cost.
+    """
+    from einops import rearrange  # noqa: F401  (imported in flux2_models)
+    from musubi_tuner.flux_2.flux2_models import (  # local to avoid cycles
+        AttentionParams,
+        timestep_embedding,
+    )
+
+    split_at: int = model._flux2_dual_gpu_split_at  # type: ignore[attr-defined]
+    cuda1 = torch.device("cuda:1")
+    cuda0 = torch.device("cuda:0")
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        x_ids: torch.Tensor,
+        timesteps: torch.Tensor,
+        ctx: torch.Tensor,
+        ctx_ids: torch.Tensor,
+        guidance: Optional[torch.Tensor],
+    ) -> torch.Tensor:
+        num_txt_tokens = ctx.shape[1]
+
+        timestep_emb = timestep_embedding(timesteps, 256)
+        del timesteps
+        vec = self.time_in(timestep_emb)
+        del timestep_emb
+        if self.use_guidance_embed:
+            guidance_emb = timestep_embedding(guidance, 256)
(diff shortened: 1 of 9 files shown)


genno-whittlery:
Heads-up: I've expanded this PR's scope from FLUX.2-only to **FLUX.2 + Wan 2.2**.

The reason is structural — the Wan 2.2 dual-GPU port edits the same two shared files this PR already touches (`networks/lora.py`'s LoRA device-pin, `hv_train_network.py`'s DDP-bypass / CPU-load / `device_placement` branch). Landing Wan as a separate PR would have guaranteed a conflict on those files for whichever merged second. The dual-GPU machinery is genuinely shared infrastructure, so one PR per trainer is the honest unit of work.

What's added on top of the original FLUX.2 changeset:
- `src/musubi_tuner/wan/wan_dual_gpu.py` (new, ~290 LOC) — the Wan `WanModel` split helper
- small hooks in `wan/modules/model.py` + `wan_train_network.py`
- the two shared files now gate on `FLUX2_DUAL_GPU` **or** `WAN_DUAL_GPU` (supersets of the prior FLUX.2-only gates; the FLUX.2 path is unchanged)

Wan 2.2 validated end-to-end on 2× RTX 5090 — T2V-A14B high-noise expert, 40 blocks split 20/20 across cuda:0/cuda:1, 10 steps, ~1.0–1.6 it/s, `--fp8_base`, loss stable. The FLUX.2 validation from the original PR description still stands unchanged.

Happy to split this back into two PRs if you'd prefer to review them separately — just say the word and I'll rebase Wan onto a fresh branch off `main`.

FurkanGozukara:
this is linux only or works on Windows as well?

genno-whittlery:
Update on the A14B dual-expert checkbox in the test plan — it's not just "untested", it's a **known limitation**, so flagging it honestly:

**Single-expert** Wan A14B (one `--dit`, no `--dit_high_noise`) — ✅ validated, works with the dual-GPU split (40 blocks distributed 20/20 across cuda:0/cuda:1, training clean).

**Dual-expert** (`--dit` + `--dit_high_noise` + `--timestep_boundary`) — ❌ does **not** work with `WAN_DUAL_GPU=true` yet. An i2v-A14B smoke trains cleanly for 14/20 steps on the first (distributed) expert, then crashes the step `--timestep_boundary` fires the high/low swap:

```
RuntimeError: Input type (struct c10::BFloat16) and bias type (struct c10::Float8_e4m3fn) should be the same
```

at `patch_embedding` (a base-model Conv3d). Root cause: `swap_high_low_weights` does `load_state_dict(assign=True)`, which (a) collapses the cuda:0/cuda:1 split back onto one device and (b) brings the incoming expert's raw weights in *bypassing the fp8-optimization exclude logic* — so base modules that should stay bf16 come back fp8. The inline LoRA device-pin only re-places LoRA params, not base modules, so it can't cover this.

A proper fix needs a re-distribution + dtype-fixup hook at the `swap_high_low_weights` site (`enable_wan_dual_gpu` is idempotent so re-invoking it is safe — the dtype mismatch also needs handling). I've documented this plainly in the `wan_dual_gpu.py` docstring (commit `2c0d18c`) and am tracking it as follow-up. The single-expert path — which is the common LoRA-training case — is unaffected.

genno-whittlery:
Follow-up: the A14B dual-expert limitation I flagged above is **now fixed** (commit `506a856`).

Root cause traced: `load_transformer` snapshots `model_high_noise.state_dict()` raw out of `load_wan_model` — the dual-GPU treatment only ran `if blocks_to_swap > 0`, and dual-GPU runs with `blocks_to_swap == 0`. So the inactive snapshot was CPU-resident, un-distributed, and carried `load_wan_model`'s raw per-module dtype layout, which diverges from the active `--dit` model (post `accelerator.prepare` + `enable_wan_dual_gpu`). `swap_high_low_weights`' `load_state_dict(assign=True)` then rebound the active model's tensors to those raw ones → the bf16/fp8 mismatch.

Fix in `swap_high_low_weights` (the `blocks_to_swap == 0` branch), gated on `is_dual_gpu_enabled()` so non-dual-GPU and FLUX.2 paths are untouched: after the `assign=True` load, cast the incoming params/buffers back to the outgoing model's dtype layout (captured from the `state_dict` snapshot taken just above), then re-invoke `enable_wan_dual_gpu` (idempotent) to restore the split + reinstall the split-aware forward. Kept the idle expert CPU-resident rather than distributing both 14B experts simultaneously (~28 GB fp8 weights across 2 cards would be too tight) — the swap pays a one-time CPU→GPU redistribute when `--timestep_boundary` fires.

**Validated on 2× RTX 5090**: i2v-A14B, `--fp8_base`, 20 steps, both experts exercised — 3 distribution events logged (1 initial + 2 swap-driven re-distributes), loss stable, checkpoints saved, `rc=0`. The same run previously crashed at step 14 on the first swap.

So the test-plan checkbox for the A14B dual-expert path can now be checked.

genno-whittlery:
Update: **HunyuanVideo** dual-GPU support folded in (commit `5cb720c`), title/description updated.

`src/musubi_tuner/hunyuan_video_dual_gpu.py` — `enable_hv_dual_gpu`. Since `hv_train_network.py` is musubi-tuner's *native* trainer and `HYVideoDiffusionTransformer` has the double-stream + single-stream structure, this port mirrors the FLUX.2 shape more than Wan: all 20 `double_blocks` + `single_blocks[:split_at]` on cuda:0, `single_blocks[split_at:]` + `final_layer` on cuda:1, single-boundary forward replacement with the boundary mid-`single_blocks`. `_bridge_block_args` recurses the per-block positional arg lists (the `freqs_cis` tuple included).

It was the cleanest of the musubi ports — RoPE is passed as forward args rather than a plain model attribute (no Wan-style `freqs` move needed), and the two text encoders are pre-cached + `del`'d before the training loop, so no text-encoder device handling. The shared-infra edits already on this branch just needed a third `HV_DUAL_GPU` clause added to each (`hv_train_network.py` DDP-skip + loading-device, `networks/lora.py` device-pin); `move_to_device_except_swap_blocks` in `hunyuan_model/models.py` early-returns through `enable_hv_dual_gpu`.

Validated on 2× RTX 5090: HunyuanVideo, `--fp8_base`, `distributed 20 double + 40 single blocks (cuda:0: 20 double + 20 single; cuda:1: 20 single + final_layer)`, 20 steps at ~1.30 it/s, avr_loss 0.155 → 0.217, checkpoint saved — **clean on the first run, no fix-iterate cycle**.

Also: the Wan 2.2 A14B dual-expert checkbox in the test plan is now ticked — that path is validated (the earlier `swap_high_low_weights` fix from commit `506a856`).

This PR now covers **FLUX.2 + Wan 2.2 + HunyuanVideo** on musubi-tuner. As always, happy to split per-trainer if you'd rather review separately.

kohya-ss:
Thank you for the very thorough write-up, the clean separation of per-trainer ports, and especially for the honest test-plan reporting (the A14B swap follow-up was a model example of how to handle a regression mid-review). The code quality here is high and the validation is convincing.

That said, I don't think this is a good fit to land in musubi-tuner upstream, and I'd like to explain the reasoning so it doesn't read as a quality judgement — it isn't.

**1. Architectural direction.** The dual-GPU path adds env-var-gated branches in shared infrastructure (`hv_train_network.py`, `networks/lora.py`, plus `move_to_device_except_swap_blocks` in three model files) and replaces each model's `forward` with a copied-and-modified variant in a side file. With three trainers covered today and more architectures likely in the future, this shape grows as N-way `or` chains in the shared files and as N parallel `forward` copies that have to be kept in sync with the canonical ones by hand. The split-aware forwards in particular are the part I'm most worried about long-term: any future change to `WanModel.forward` / `HYVideoDiffusionTransformer.forward` / `Flux2.forward` would silently drift from the split copies, and there's no test surface that would catch it.

**2. Scope vs. user base.** musubi-tuner's core focus is memory-efficient *single-GPU* LoRA training (fp8 + block-swap  + gradient checkpointing). I do recognize the real gap your PR addresses: DDP today isn't compatible with `--blocks_to_swap` (plus DDP itself is flaky on Windows), so users who *do* have multiple GPUs but not enough VRAM per card don't have a good story here. That said, the population of users with two or more 24 GB+ cards is fairly small, so the addressable audience for this path stays narrow in practice.

**Suggested path.** Given that you already maintain https://github.com/genno-whittlery/flux2-dual-gpu-lora as a cross-trainer hub for exactly this pattern (ai-toolkit, OneTrainer, DiffSynth-Studio, and now musubi-tuner), I'd like to suggest keeping the musubi-tuner port living there as a patch/fork rather than in upstream. That keeps the dual-GPU  machinery in one place under your maintenance, lets you iterate quickly without an upstream review loop, and lets musubi-tuner upstream stay focused on the single-GPU path. If you wish, I'm happy to link your repo from the musubi-tuner README's "related projects" section so users who need this path can find it.

**Suggested path.** Given that you already maintain https://github.com/genno-whittlery/flux2-dual-gpu-lora as a cross-trainer hub for exactly this pattern (ai-toolkit, OneTrainer, DiffSynth-Studio, and now musubi-tuner), I'd like to suggest keeping the musubi-tuner port living there as a patch/fork rather than in upstream. That keeps the dual-GPU  machinery in one place under your maintenance, lets you iterate quickly without an upstream review loop, and lets musubi-tuner upstream stay focused on the single-GPU path. For discoverability, would you consider opening a Discussion topic on this repo pointing to your hub? A Discussion is the most natural home for users searching for the dual-GPU path, and it stays under your control if anything about the hub repo changes over time.

I'll close this PR with that in mind, but please don't take it as the work being wasted — it's clearly useful to the users who need it, and the right home for it is your hub repo. Thanks again for the careful work and the transparent reporting.

PR #956, opened 2026-06-02
Fused QKNorm + RoPE and packed QKV

rockerBOO:
For Flux2 we have QKNorm which requires QK "separated" and the rest doesn't need QKV separated. 

This PR works to keep QKV together all the way through attention (specifically flash attention) using a fused QKNorm + RoPE triton kernel. 

Getting ~7.8% improved performance and increased tensor usage. Doesn't need to allocate new tensors on the CPU and then transfer them to the GPU. We do the QKNorm in place. 

There is a benchmark  in here which showing around 2.4x to 5.8x (Edit: now 3.8x to 10x) as fast as the reference implementation. 

This is my first attempt at a kernel so will probably need edge cases to be covered, and maybe it should be opt-in instead of always on. Also has a reference implementation that works without the fusion. 

Specifically posting this to get feedback if this is desired, and any concerns about this implementation.

References:

https://www.kapilsharma.dev/posts/triton-kernels-rms-norm/
https://github.com/nvxuanyuc/TensorRT-LLM/blob/main/cpp/tensorrt_llm/kernels/fusedQKNormRopeKernel.cu

### DIFF
--- a/notes/2026-06-10-triton-reference-f32.md
+++ b/notes/2026-06-10-triton-reference-f32.md
@@ -1,0 +1,92 @@
+# Triton fused norm+rope: reference path diverged from original (2026-06-10)
+
+Branch: `packed`. Files: `src/musubi_tuner/kernels/fused_norm_rope.py`,
+`tests/test_packed_vs_original.py`, `tests/test_fused_norm_rope.py`,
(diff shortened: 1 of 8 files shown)


rockerBOO:
Closing in favor of #968

PR #966, opened 2026-06-11
Add Ideogram 4 FP8 support

sdbds:
## Summary
- Add Ideogram 4 FP8 support with local Comfy-Org single-file component loading and automatic official tokenizer/config download.
- Add Ideogram 4 latent/text cache helpers, sampler presets, VAE/DiT utilities, asymmetric CFG generation path, and LoRA support for FP8 linear layers.
- Merge the latest upstream main and keep the upstream text-cache dtype replacement optimization for Ideogram 4 caches.

## Test Plan
- `python -m unittest discover tests -p "test_ideogram4*.py"`
- `python -m compileall src/musubi_tuner/ideogram4 src/musubi_tuner/ideogram4_generate_image.py src/musubi_tuner/ideogram4_cache_latents.py src/musubi_tuner/ideogram4_cache_text_encoder_outputs.py src/musubi_tuner/ideogram4_train_network.py src/musubi_tuner/networks/lora_ideogram4.py src/musubi_tuner/dataset/cache_io.py tests/test_ideogram4_synthetic.py`
- `ruff check src/musubi_tuner/ideogram4 src/musubi_tuner/ideogram4_generate_image.py src/musubi_tuner/ideogram4_cache_latents.py src/musubi_tuner/ideogram4_cache_text_encoder_outputs.py src/musubi_tuner/ideogram4_train_network.py src/musubi_tuner/networks/lora.py src/musubi_tuner/networks/lora_ideogram4.py src/musubi_tuner/dataset/cache_io.py tests/test_ideogram4_synthetic.py`
- `git diff --check upstream/main...HEAD`

image

PR #975, opened 2026-06-18
Ideogram4 followup

kohya-ss:
Follow-up for #966.

- Improve numerical accuracy of MRoPE in bfloat16. This enables bf16 mixed precision training.
- Enable mixed precision (autocast) training.
- Remove Fp8Linear by utilizing existing code (support pre-quantized weights).
- Swap text-image order to support the unified attention (supports FlashAttention, SDPA without  mask etc.)
- Support updated block swap (H2D only etc.)


amrnidal999-tech:
Hey kohya, still havent tried these followups

Based on my testing after u first merged seemed like it learns but it gets to a point early on aswell not late.

The image starts becoming complete noise. Adma8bit 2e-5 around 3k steps pure noise Will test this out today thank you!

kohya-ss:
Ideogram4 seems to be a rather unique model, and I haven't been able to train it successfully yet.

This PR should be mathematically the same as #966, but if you encounter any problems, try training it with `--mixed_precision no`. Specifying mixed_precision should produce slightly different results.

amrnidal999-tech:
> Ideogram4 seems to be a rather unique model, and I haven't been able to train it successfully yet.
> 
> This PR should be mathematically the same as #966, but if you encounter any problems, try training it with `--mixed_precision no`. Specifying mixed_precision should produce slightly different results.

Well so few tests with this to confirm later on.

it is seems to be a harder model to train but based on my testing.

adamw8bit 2e-4 completely blows it out

it takes time learn characters has good understanding of faces rather than bodies so bodies take more time to train.

automagic v3 on ai toolkit lokr full rank factor of 4 learned the face really well body at 10k steps as if it was not there- i suspect this is a factorization issue could need factor 2.

i did adamw8bit 1e-4 still learns very fast but wanted to get body likeness- so pushed Lora rank to 96 and finally seems like i got something but I overfried it with a 30img dataset.  Even at 1e-4 face was almost fully learned by 1.2k steps and kind of getting overbaked.

pushing dataset now to 85 images and will do testing on AI toolkit before i come back to here.

kohya-ss:
Thanks for the review @Copilot!

### Comment 1 (`fp8_optimization_utils.py`) — 1-byte dtype gating

Good catch — applied. The accept path now only keeps weights as-is when the dtype is an actual float8 (`torch.float8_e4m3fn` / `torch.float8_e5m2`); other 1-byte dtypes such as uint8/int8 still raise as before, so they can't slip into the FP8 monkey-patched forward and crash on `torch.finfo`.

```python
if original_dtype.itemsize == 1:
    if allow_prequantized_fp8 and original_dtype in (torch.float8_e4m3fn, torch.float8_e5m2):
        ...
```

### Comment 2 (`ideogram4_generate_image.py`) — `--lora_weight` nargs

The observation is correct, but I'll leave it as `nargs="*"` for consistency. Every other inference script in this repo (`wan` / `flux_2` / `qwen_image` / `hv` / `fpack` / `hidream_o1` / `kandinsky5` / `zimage`, …) declares `--lora_weight` with `nargs="*"`, and the empty-list-skips-LoRA behavior is shared across all of them. Switching only this script to `nargs="+"` would diverge from that convention; if we want to tighten it, it should be a repo-wide change rather than something specific to Ideogram 4.

PR #982, opened 2026-06-23
Add Krea 2 model support and update version to 0.3.4

kohya-ss:
This pull request introduces experimental support for the Krea 2 architecture (LoRA training and inference) to the project. The main changes include updates to the documentation in both English and Japanese, as well as the addition of a new dedicated documentation file for Krea 2. Additionally, a new script entry point for Krea 2 latent pre-caching has been added.

**Krea 2 Integration and Documentation:**

* Added experimental support for Krea 2 (LoRA training and inference), as announced in both `README.md` and `README.ja.md`, with references to the new documentation and the relevant pull request. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R66-R69) [[2]](diffhunk://#diff-b3fc642ee9795d9cb415501fb1ca7858d4222eb6f32ae8dc26ae29b3f1934336R61-R64)
* Added comprehensive documentation for Krea 2 usage in `docs/krea2.md`, including model requirements, pre-caching, training, inference, memory optimization, and LoRA configuration, with bilingual (English/Japanese) instructions.


sdbds:
I think the turbo‑cache-dit approach is a bit too heavy.
Could we instead use a reverse merge, leveraging （turbo‑raw） to generate a LoRA that can be simply integrated into the raw inference sampling?
like 8steps lora.

kohya-ss:
It is certainly computationally heavy to handle both models, whether using a caching approach or loading them on the fly. 

I have already tried the difference Turbo LoRA, but the LoRA extracted from the delta didn't perform very well—even at rank=64—as the resulting images differed from the original Turbo. For verification purposes, though, running inference with the raw model at 8 steps might be a viable option.

sdbds:
I think maybe we should try setting the rank to 1. Just get the acceleration effect and then use raw 8-step lora.

PR #987, opened 2026-07-01
feat: wavelet-domain auxiliary loss trainer for FLUX.2

rockerBOO:
## Summary

Adds `flux_2_train_network_wavelet_loss.py`, a FLUX.2 LoRA training entry point that augments the standard flow-matching MSE with a frequency-domain auxiliary term: `loss = mse.mean() + alpha * wavelet_loss`. The wavelet term is computed by the optional external [wavelet-loss](https://github.com/rockerBOO/wavelet-loss) package (DWT/SWT with PyWavelets-parity transforms), weighting low/high-frequency bands separately so the model gets an explicit signal on coarse structure vs fine detail.

**Zero modifications to `flux2_models.py` or any base trainer** — the trainer is a subclass of `Flux2NetworkTrainer` implemented entirely on existing extension seams:

- `call_dit`: stashes `noisy_model_input` on `DiTOutput.extra`
- `on_train_start`: constructs the `WaveletLoss` module on the training device
- `compute_loss`: adds the wavelet term to the weighted MSE
- `extra_metadata`: embeds the full wavelet config (`ss_wavelet_loss_*`) in saved safetensors

### Loss space

By default (`--wavelet_loss_rectified_flow`) the wavelet term runs on estimated clean latents recovered from the velocity prediction via the flow-matching identity (`x0 = noisy - sigma * v`), so it penalises the frequency structure of the estimated clean image and reaches zero at a perfect prediction. `--no-wavelet_loss_rectified_flow` runs it directly on raw velocity prediction vs target (AWWL-style), dropping the implicit sigma² weighting.

### Timestep weighting

The wavelet term is faded by a sigmoid over the timestep — ~1 at low noise (where high-frequency detail is meaningful) fading to ~0 near pure noise — configurable via `--wavelet_loss_timestep_cutoff` / `--wavelet_loss_timestep_transition_width` / `--wavelet_loss_max_timestep` (default 1000 to match the trainer's scheduler-footing timesteps).

### Optional dependency

The wavelet term requires the `wavelet-loss` package, installed into the same environment:

```bash
uv pip install wavelet-loss
# or: pip install wavelet-loss
```

It is imported behind a guard: the script raises a clear `ImportError` only when `--wavelet_loss` is actually requested. Nothing else in the repo depends on it, and it is not added to `pyproject.toml` dependencies.

## Validation

- 19 CPU tests in `tests/test_wavelet_loss_trainer.py` covering arg parsing, x0 recovery (target recovers latents exactly), compute_loss combination (disabled path equals plain weighted MSE; enabled path changes the loss and returns `wavelet_loss/` metrics), continuous off-schedule timesteps, and metadata. Package-dependent tests `importorskip` cleanly when wavelet-loss is not installed.
- Training runs on FLUX.2 with both loss spaces; metrics stratified by timestep weight to avoid sampling confounds.

## Notes for review

- `pyproject.toml` gains `[tool.pytest.ini_options] pythonpath = ["src"]` so the test suite runs from a clean checkout (same note as #969; whichever merges second resolves trivially).
- Internal extension entry point — no API stability guarantees.

## Usage

```bash
uv pip install wavelet-loss

accelerate launch --num_cpu_threads_per_process 1 --mixed_precision bf16 \
  src/musubi_tuner/flux_2_train_network_wavelet_loss.py \
  --dit flux-2-klein-base-4b.safetensors ... \
  --wavelet_loss --wavelet_loss_alpha 0.1 \
  --wavelet_loss_transform swt --wavelet_loss_wavelet sym7 --wavelet_loss_level 2
```

---

## 概要（日本語）

FLUX.2 LoRA 学習用のエントリポイント `flux_2_train_network_wavelet_loss.py` を追加します。標準の flow-matching MSE に周波数領域の補助損失を加算します（`loss = mse.mean() + alpha * wavelet_loss`）。ウェーブレット項はオプションの外部パッケージ [wavelet-loss](https://github.com/rockerBOO/wavelet-loss)（PyWavelets と一致する DWT/SWT 変換）で計算され、低周波・高周波バンドを個別に重み付けします。

**`flux2_models.py` やベーストレーナーへの変更は一切ありません** — `Flux2NetworkTrainer` のサブクラスとして、既存の拡張シーム（`call_dit` / `on_train_start` / `compute_loss` / `extra_metadata`）のみで実装しています。

- デフォルト（`--wavelet_loss_rectified_flow`）では flow-matching の恒等式 `x0 = noisy - sigma * v` で復元した推定クリーン潜在にウェーブレット項を適用します。`--no-wavelet_loss_rectified_flow` で velocity 空間に直接適用（AWWL 方式）。
- ウェーブレット項はタイムステップに応じてシグモイドでフェードします（低ノイズで ~1、純ノイズ付近で ~0。`--wavelet_loss_timestep_cutoff` 等で設定可能）。
- 使用するには `uv pip install wavelet-loss`（または `pip install wavelet-loss`）でパッケージをインストールしてください。ガード付きインポートのため、`--wavelet_loss` 指定時のみ明確な `ImportError` を出します。`pyproject.toml` の依存関係には追加していません。
- CPU テスト 19 件（パッケージ未インストール時は skip）。
- `pyproject.toml` に `[tool.pytest.ini_options] pythonpath = ["src"]` を追加（#969 と同一の変更）。
- 内部拡張ポイントのため API 安定性の保証はありません。

AI-assisted


recris:
I've been privately using a haar wavelet loss estimator for training Flux.2 and Krea2 and it works remarkably well, better than straight MSE loss.

However I think the approach to extension seams, as currently designed, is not the best fit for integration non-standard loss estimators. Every new loss function needs a subclass on every type of network trainer, and this is means I can't have a generic loss implementation reusable across models, plus it is not easy to combine loss functions, it always needs ad-hoc code for every model and combination of loss functions.

To me, loss estimation should be pluggable into the trainer in a way similar to how optimizers are configurable - we configure an estimator class (or multiple classes) with a well defined interface, it then can be reused at will via configuration.

I might submit a more complete proposal as an issue/PR on how it could be done (if get the time).

rockerBOO:
So your argument here is it shouldn't be a seam implementation but a separate loss function built down in the main trainer code? 

So you're saying we make the loss function something we import by passing the path

```toml
loss_fn = "wavelet_loss.musubi.WaveletLoss"
```

Note: will match the names up with what is related in musubi-tuner or sd-scripts

only issue is how do we setup the parameters for the loss if we do a function compared to a torch.nn.Module class method. this would add in something like

```toml
loss_fn_args = ["transform=swt"] 
```

I think i like this format better because the issue was a bit of the in the trainer code and also for experimentation and trying new loss functions. 

Some newer methods like PO would have loss functions that include loading new models, which was done for a model recently to add DiNO but this would allow us to experiment with that and other options. I still want to add back in huber loss and other loss functions from sd-scripts but might be best with built in loss functions (and they are exposed via `loss_fn`)

PR #988, opened 2026-07-02
feat: add --log_grad_metrics for pre-clip gradient norm logging

rockerBOO:
## Summary

Adds an opt-in `--log_grad_metrics` flag that logs gradient diagnostics to the tracker (TensorBoard/WandB) each optimizer step, for all `*_train_network.py` trainers via `NetworkTrainer`:

- `grad/norm` — total L2 norm across all network gradients (**pre-clipping**, so it shows what `--max_grad_norm` is actually fighting)
- `grad/mean_norm` — mean of per-parameter L2 norms
- `grad/max` — max absolute gradient element

Useful for spotting gradient explosions/vanishing and for choosing a sensible `--max_grad_norm` value.

## Implementation notes

- Computed after `accelerator.backward()` and the manual DDP grad reduce, before `accelerator.clip_grad_norm_`, so the logged norm is the true pre-clip value.
- Metrics are emitted only on gradient-sync steps (i.e. actual optimizer updates, gradient-accumulation aware); accumulation micro-steps log nothing.
- Default **off**: the `.item()` calls force a GPU sync per step, so the flag is opt-in and additionally gated on an active tracker (`len(accelerator.trackers) > 0`). With the flag off there is no behavior or performance change.
- No per-trainer changes — everything lives in `training/trainer_base.py` and `training/parser_common.py`.

## Tests

`tests/test_grad_metrics.py`: 9 pure-CPU tests covering the norm math (single/multi-param total norm, mean-vs-total distinction, max element, params without grads, empty case) and the parser flag default.

```
uv run pytest tests/  # 85 passed
```

### DIFF
--- a/src/musubi_tuner/training/parser_common.py
+++ b/src/musubi_tuner/training/parser_common.py
@@ -253,6 +253,13 @@ def _add_logging_args(parser: argparse.ArgumentParser) -> None:
         help="specify WandB API key to log in before starting training (optional). / WandB APIキーを指定して学習開始前にログインする（オプション）",
     )
     parser.add_argument("--log_config", action="store_true", help="log training configuration / 学習設定をログに出力する")
+    parser.add_argument(
+        "--log_grad_metrics",
+        action="store_true",
+        help="log gradient norm metrics (grad/norm, grad/mean_norm, grad/max, pre-clipping) to the tracker."
+        " Adds a small per-step GPU sync overhead"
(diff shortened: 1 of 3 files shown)


kohya-ss:
Thanks for this, and sorry for the delay in getting to it!

The implementation looks great — collecting the metrics after the DDP reduce and before clipping is exactly right, the tracker guard keeps it cost-free when unused, and I appreciate the tests. Naming (`grad/*`) fits nicely with the existing `loss/*` keys too.

I'll merge this as-is and add the documentation on our side in a follow-up. Thanks again for the clean, well-scoped contribution!

PR #992, opened 2026-07-06
Add custom loss function support

rockerBOO:
This allows us to inject our own custom loss function in. Takes the same signature as compute_loss. 

Would want some feedback on this approach as it feels a bit opinionated towards a ctx object with the information from the compute_loss function. This approach makes it easier to add new options without breaking previous versions and a cleaner interface but it's more abstract from just passing the parameters over.

This idea is credit to @recris who suggested with the wavelet loss implementation in #987. 

Related #987

### DIFF
--- a/src/musubi_tuner/training/loss.py
+++ b/src/musubi_tuner/training/loss.py
@@ -21,11 +21,14 @@ import ast
 import functools
 import importlib
 import inspect
+import logging
 from typing import Any, Callable, Optional
 
 import torch
 
 from musubi_tuner.training.timesteps import compute_loss_weighting_for_sd3
+
+logger = logging.getLogger(__name__)
 
 
 def mse_loss(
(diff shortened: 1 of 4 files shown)


rockerBOO:
I have been using this extensively and works as expected. Just some UX of how we are using it might need to be discussed some more. 

As an example of usage. 

```toml
loss_fn = "wavelet_loss.musubi.WaveletPlusSNRHuber"
loss_fn_args = [
  "alpha=1.0", 
  "loss_type='snr_huber'", 
  "snr_huber_cmin=0.2",
  "snr_huber_cmax=1.0",
  "transform_type='swt'", 
  "wavelet='sym7'", 
  "level=2",
  "band_weights={'ll': 0.0, 'lh': 1.0, 'hl': 1.0, 'hh': 1.0}",
  "ll_level_threshold=2", 
  "metrics=True", 
  "rectified_flow=False",
```

The Loss function must match our format, thats the tradeoff here with flexibility. 

Example: https://github.com/rockerBOO/wavelet-loss/blob/ef12c2d6784b793f8a71b207125f01e1c34f76b4/src/wavelet_loss/musubi.py#L159-L220

PR #999, opened 2026-07-12
Add portable native toolchain setup for torch.compile

FurkanGozukara:
## Summary

- Add portable native-toolchain discovery and validation for PyTorch Inductor on Windows and Linux.
- Prepare the environment before model loading when direct `--compile` or Accelerate `--dynamo_backend inductor` is requested.
- Validate the selected compiler by compiling, linking, and running real probe programs instead of relying only on `PATH` checks.
- Add stable Inductor/Triton caches, a diagnostic CLI, and a guarded eager fallback for cold compilation failures.
- Leave non-Inductor and non-compile training paths unchanged.

## Problem

Musubi supports `torch.compile`, but native Inductor compilation depends on host state that varies substantially between installations:

- On Windows, `cl.exe` may be visible while `INCLUDE`, `LIB`, or the Windows SDK environment is missing.
- Multiple Visual Studio editions and MSVC toolsets may be installed, and the newest toolset is not always accepted by the installed CUDA Toolkit.
- CUDA environment variables can point at a toolkit that does not match `torch.version.cuda`.
- Visual Studio Build Tools may live on a custom drive or be discoverable only through `vswhere` or the registry.
- Linux systems can expose several GCC, Clang, NVIDIA HPC, Intel, ARM, or Conda compilers with different CUDA compatibility.
- Ninja and compile-cache locations differ across virtual environments and launch methods.

These problems otherwise surface during the first forward pass, after model loading, with compiler errors that are difficult for users to diagnose. GUI launchers can prepare a parent environment, but direct CLI training and Accelerate worker processes also need a self-contained setup inside Musubi.

### Portable discovery package

A new dependency-light `musubi_tuner.torch_compile_toolchain` package provides:

- `ensure_compile_environment` for the current process.
- `prepare_compile_subprocess_env` for launchers.
- Structured `CompileToolchainStatus` diagnostics.
- `compile_callable` / `compile_module_callable` wrappers with first-call verification and eager fallback.
- `python -m musubi_tuner.torch_compile_toolchain` diagnostics and a real `--smoke-test`.

The discovery code uses the Python standard library until it needs to inspect PyTorch's CUDA version.

### Windows

Windows discovery covers:

- Active Visual Studio developer environments.
- Full Visual Studio and standalone Build Tools installations.
- `vswhere`, registry entries, standard paths, custom drives, prerelease, and legacy layouts.
- Explicit `MUSUBI_VS_INSTALLDIR`, `MUSUBI_VS_DEV_CMD`, and `MUSUBI_VSWHERE` overrides.
- Every installed x64 MSVC toolset, ordered by its actual toolset version.
- Exact and family `-vcvars_ver` selection with fallback across candidates.

A candidate is accepted only after it can:

1. Compile, link, and run a C++17 program using standard-library headers.
2. Compile and link OpenMP when required.
3. Compile a CUDA source with `nvcc` when a CUDA Toolkit is available.

This rejects the common partial state where `cl.exe` is on `PATH` but headers, libraries, or the SDK are unavailable.

### Linux

Linux discovery honors explicit `NVCC_CCBIN`, `CUDAHOSTCXX`, `CXX`, and `CC` settings, then checks common versioned GCC/Clang paths plus NVIDIA HPC, Intel, ARM, and Conda candidates. Each candidate runs the native C++ probe and, when available, an `nvcc -ccbin` compatibility probe.

### CUDA, Ninja, and caches

- CUDA candidates come from explicit variables, Conda, standard installations, and `nvcc` on `PATH`.
- An exact `torch.version.cuda` match is preferred, followed by the nearest minor in the same major release.
- Ninja is discovered from overrides, `PATH`, virtual environments, Conda, the Python package, and Visual Studio CMake installations.
- Stable project-local Inductor and Triton cache directories are configured, with a writable temporary fallback.
- `.cache/` is ignored so successful compilation does not dirty the repository.

### Training integration

- The shared training accelerator setup and the legacy Hunyuan Video accelerator path prepare the native toolchain before constructing `Accelerator` when Inductor is requested.
- `compile_transformer` verifies the environment before wrapping blocks with the Inductor backend.
- The first invocation of each compiled block is guarded. A cold compiler failure disables compiled execution for the remaining unverified blocks and continues eagerly by default.
- Failures after a block has already compiled and executed successfully are re-raised rather than hidden.
- `MUSUBI_TORCH_COMPILE_FALLBACK=0` disables the first-call fallback when strict failure behavior is desired.

Non-native backends such as `eager` do not trigger toolchain discovery.

## Real Windows validation

Environment:

- Windows, Python 3.12.10
- PyTorch 2.13.0 + CUDA 13.0
- RTX 5090
- Visual Studio 2022 Community and multiple local CUDA/toolchain paths

The diagnostic selected and validated:

- CUDA Toolkit 13.1 as the nearest compatible CUDA 13.x installation
- MSVC 14.44.35207 through `VsDevCmd.bat`
- C++17 compile, link, and execution
- OpenMP compile and link
- CUDA host-compiler compatibility through a real `nvcc` probe
- Ninja discovery

Both real compile paths succeeded:

```text
torch.compile smoke test: passed (cuda, loss=0.490634)
setup_ready: compiled wrapper created
first_call_ok: first compiled call succeeded
```

The standalone module example also compiled and executed a CUDA model successfully.

## Tests

```text
40 passed, 2 subtests passed
```

The test coverage includes:

- CUDA version selection and stale-environment replacement.
- Visual Studio roots, custom paths, toolset ordering, and setup-script environment capture.
- Incomplete `cl.exe` environments and fallback candidate selection.
- C++17, OpenMP, and CUDA compiler probes.
- Linux compiler selection and fallback.
- Ninja and cache discovery.
- Direct compile and Accelerate Dynamo request detection.
- Training entry-point integration.
- Safe compile wrappers and first-call eager fallback.
- State-dict behavior around compiled block wrappers.

Additional checks:

- `ruff check` passed for every changed Python file.
- `ruff format --check` passed for every changed Python file.
- Existing Ideogram4 FP8 and Krea 2 focused tests remained green.

Linux discovery is covered through isolated platform/candidate tests; the real native smoke test reported above was run on Windows.


kohya-ss:
Thank you very much for this PR, and for the amount of care that clearly went into it — the discovery logic, the real compile/link probes, and the test coverage are all impressive.

After going through it in detail, though, I've decided not to merge it. Let me explain my reasoning, since it's about scope rather than code quality.

**What the PR effectively does**
The toolchain is discovered and activated, but nothing is auto-installed. So the parts of the Windows setup that are actually painful — installing the MSVC Build Tools and installing Triton — are unchanged; users still have to do those by hand. In practice, the net benefit is automating the `vcvars64.bat` / "x64 Native Tools Command Prompt" step for people who have already installed everything.

**It's gated too broadly, and it can break a currently-working case**
On Windows, MSVC is only required for `--compile_dynamic=true`; plain `--compile` works with Triton alone (this is what our docs state as well). But the training hook fires for any `--compile` + inductor run regardless of `compile_dynamic`, and it raises/aborts when the MSVC environment isn't found (with `require_openmp=True` by default). That means a user who today runs plain `--compile` with only Triton installed (no Build Tools) would be blocked before training even starts. So for the most common Windows case it's a regression, in exchange for automating one manual step for the narrower `--compile_dynamic` case.

**Cost vs. benefit**
Weighing that fairly narrow benefit against ~2,700 lines of intricate, platform-specific code that I would need to maintain on my own indefinitely (plus a public API surface that isn't used by the training path yet), I don't think the trade-off works for this project right now.

If we ever want to smooth over the `vcvars` step, I think a much lighter approach fits better — e.g. a small launcher batch file that calls `vcvars64.bat` and then `accelerate launch`, which covers the same need without adding a large subsystem to the codebase.

I really appreciate the effort here, and I'm sorry to close it after all the work you put in. Please don't let this discourage you from contributing again — thank you!

FurkanGozukara:
sure thanks fore reply. currently works perfect with my app :D really improves speed

PR #1000, opened 2026-07-12
Fix training progress rate after the first warmup step

FurkanGozukara:
## Problem

Training progress uses `tqdm(..., smoothing=0)`, so its displayed rate is calculated from `(n - initial) / elapsed`.

The existing warmup exclusion resets the bar **before** the first `update(1)`. That restarts the clock, but leaves the baseline at zero. At step 4, for example, the elapsed time covers only steps 2-4 while the rate calculation still counts all four steps. This is why the reported `3.61s/it` initially looks substantially faster than the stable `~4.7s/it` rate and produces an optimistic ETA.

## Solution

Use the first completed synchronized training step as the timing baseline:

- update the visible bar to step 1 normally
- synchronize through `loss.detach().item()`, ensuring queued CUDA work from the first step is complete
- preserve the visible step and total while setting tqdm's rate baseline to the current count
- restart cumulative timing and clear EMA timing history
- refresh the first-step display to an unknown rate; step 2 and later then use only post-warmup timings
- no-op for disabled progress bars on non-main distributed processes

The helper uses tqdm's own clock, so there is no platform-specific timer behavior.

## Training paths covered

The shared behavior is applied to every training progress loop currently in the repository:

- general `NetworkTrainer` model training
- legacy Hunyuan Video training
- Qwen Image training
- Z-Image training
- HiDream O1 training

It runs only on synchronized gradient steps, so gradient accumulation does not reset the timer early.

## Result

The visible count remains correct at `1 / total`, but the first step no longer contaminates rate or ETA. Starting when step 2 completes, both values are calculated from completed post-warmup steps only.

A live wall-clock check used a 2-second first step followed by three 1-second steps:

```text
n=4 baseline=1 measured_steps=3 elapsed=3.002s seconds_per_step=1.001
```

## Validation

- `python -m pytest -q`: **64 passed**
- focused timing tests cover cumulative rate/ETA, EMA reset, disabled bars, and all five loop integrations
- `python -m ruff check`: passed for all touched files
- `python -m ruff format --check`: passed for all touched files
- live wall-clock behavior validated in the project venv on Windows
- implementation contains no OS-specific code and follows tqdm's same clock path on Linux

### DIFF
--- a/src/musubi_tuner/hidream_o1_train.py
+++ b/src/musubi_tuner/hidream_o1_train.py
@@ -527,8 +527,6 @@ class HiDreamO1Trainer(HiDreamO1NetworkTrainer):
                 keys_scaled, mean_norm, maximum_norm = None, None, None
 
                 if accelerator.sync_gradients:
-                    if global_step == 0:
-                        progress_bar.reset()
                     progress_bar.update(1)
                     global_step += 1
 
@@ -562,6 +560,8 @@ class HiDreamO1Trainer(HiDreamO1NetworkTrainer):
                         optimizer_train_fn()
 
                 current_loss = loss.detach().item()
+                if accelerator.sync_gradients and global_step == 1:
+                    train_utils.reset_progress_bar_timing(progress_bar)
                 loss_recorder.add(epoch=epoch, step=step, loss=current_loss)
                 avr_loss = loss_recorder.moving_average
                 progress_bar.set_postfix(**{"avr_loss": avr_loss})
--- a/src/musubi_tuner/hv_train.py
+++ b/src/musubi_tuner/hv_train.py
@@ -1136,6 +1136,8 @@ class FineTuningTrainer:
                     optimizer_train_fn()
 
                 current_loss = loss.detach().item()
+                if accelerator.sync_gradients and global_step == 1:
+                    train_utils.reset_progress_bar_timing(progress_bar)
                 loss_recorder.add(epoch=epoch, step=step, loss=current_loss)
                 avr_loss: float = loss_recorder.moving_average
                 logs = {"avr_loss": avr_loss}  # , "lr": lr_scheduler.get_last_lr()[0]}
--- a/src/musubi_tuner/qwen_image_train.py
+++ b/src/musubi_tuner/qwen_image_train.py
@@ -652,8 +652,6 @@ class QwenImageTrainer(QwenImageNetworkTrainer):
 
                 # Checks if the accelerator has performed an optimization step behind the scenes
                 if accelerator.sync_gradients:
-                    if global_step == 0:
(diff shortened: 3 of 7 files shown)


kohya-ss:
Thanks for this fix, and for the thorough tests! The analysis is correct — the old code was a bit subtle: `progress_bar.reset()` zeroed `n` but left `initial` at `0`, so after excluding the first step tqdm effectively computed `rate = n / elapsed` instead of `(n - 1) / elapsed`. That off-by-one is why the rate/ETA drifted as training went on. Setting `initial` to the current step and rebaselining the start time is the right way to solve it. 👍

One thing I'd like to simplify is the dependency on tqdm's private internals — specifically the `_ema_dn` / `_ema_dt` / `_ema_miniters` handling.

All of the training loops create the bar with `smoothing=0`:

```python
progress_bar = tqdm(range(args.max_train_steps), smoothing=0, disable=..., desc="steps")
```

Under `smoothing=0`, tqdm never touches the EMAs. In `update()` the EMA is only fed when `if self.smoothing and dt and dn:`, and `format_dict["rate"]` is `self._ema_dn() / self._ema_dt() if self._ema_dt() else None`, which becomes `None` — so `format_meter` falls back to the plain `(n - initial) / elapsed`. In other words, for this codebase the EMA reset never affects the displayed rate; it's only exercised by the `smoothing=0.3` unit test.

Since we don't need to reset the EMAs ourselves, we can delegate that (plus the `start_t` / `last_print_t` reset) to tqdm's public `reset()` and only restore the visible count afterwards. That shrinks the internal surface to just `initial` / `n` / `last_print_n` (all of which are part of `format_dict`):

```python
def reset_progress_bar_timing(progress_bar) -> None:
    """Exclude the first completed step from tqdm rate and ETA calculations."""
    if getattr(progress_bar, "disable", False):
        return

    completed = progress_bar.n
    progress_bar.reset()  # public API: refreshes start_t/last_print_t and the EMAs
    progress_bar.initial = completed
    progress_bar.n = completed
    progress_bar.last_print_n = completed
    progress_bar.refresh()
```

I verified this produces identical rate/ETA output to your version for `smoothing=0` (constant `4.00s/it`, ETA counting down correctly), and it also passes all four tests you added, including `test_timing_reset_clears_exponential_average_history` — `reset()` rebuilds the EMA objects so their `.calls` are back to `0`.

The only minor trade-off is that `reset()` internally refreshes once at `n = 0` before we restore the count, so there can be a one-frame flicker on that single step. In practice it runs exactly once per training run and is usually throttled away by `mininterval`, so I think it's a fine trade for dropping the private-attribute access.

If you'd prefer, I'm happy to apply this tweak in a follow-up after merging — either way works. Thanks again for improving this!

FurkanGozukara:
sure either way fine for me thanks

PR #1003, opened 2026-07-13
Unify guarded SDPA acceleration across model trainers

FurkanGozukara:
## Summary

- Centralize SDPA backend selection for every current network and full-training entry point instead of mutating Krea 2 arguments.
- Keep PyTorch SDPA as the portable user-facing choice and prefer its native fused FlashAttention backend when available.
- When native fused SDPA is unavailable, select external FlashAttention only after fixed-length and variable-length BF16 GQA forward/backward CUDA probes pass.
- Re-check the real Q/K/V tensors at runtime and fall back to PyTorch SDPA for incompatible devices, dtypes, head dimensions, or GQA layouts while preserving padding masks.
- Add `--use_legacy_sdpa` as an explicit compatibility/benchmark opt-out. `MUSUBI_DISABLE_EXTERNAL_FLASH_SDPA=1` remains available for diagnostics.

## Coverage

The shared resolver is used by:

- the common `NetworkTrainer` path (Krea 2, Qwen Image, Wan, FLUX.2/Klein, Z-Image, Ideogram 4, and other network trainers)
- Qwen Image full training
- Z-Image full training
- the legacy HunyuanVideo full trainer

Wan and HunyuanVideo retain only small layout adapters for the internal automatic mode. Explicit FlashAttention, FlashAttention 3, SageAttention, and xFormers selections are unchanged.

## Why

Some Windows PyTorch builds expose SDPA without a native fused FlashAttention kernel. Those builds can fall back to a substantially slower memory-efficient kernel even when the installed external FlashAttention package supports the model. The previous PR handled only Krea 2 by rewriting its attention arguments. This version makes the policy shared, keeps user intent intact, and adds a per-tensor fallback needed by models such as Ideogram 4.

## Controlled Windows benchmarks

RTX 5090, Torch 2.13.0+cu130, batch size 1, rank 128 LoRA, gradient checkpointing, 20 steps, `torch.compile` disabled, and `blocks_to_swap = 0` in every pair:

| Model | Legacy PyTorch SDPA | Automatic | Change |
| --- | ---: | ---: | ---: |
| Krea 2, 1024 px | 3.50 s/it | 2.81 s/it | 19.71% faster |
| Qwen Image, 512 px | 3.05 s/it | 2.80 s/it | 8.20% faster |
| FLUX Klein 4B, 512 px | 5.10 s/it | 4.99 s/it | 2.16% faster |
| FLUX Klein 9B, 512 px | 5.79 s/it | 5.64 s/it | 2.59% faster |
| Wan 2.1 14B, 512 px | 1.58 s/it | 1.54 s/it | 2.53% faster |
| Z-Image, 512 px | 5.57 s/it | 5.42 s/it | 2.69% faster |
| Ideogram 4, 512 px | 5.79 s/it | 5.73 s/it | safe PyTorch fallback |

The Ideogram run produced FlashAttention-incompatible mixed/FP32 attention tensors. Automatic mode detected this and completed on PyTorch SDPA; its 1% difference is treated as measurement noise rather than an acceleration claim.

## Safety

Automatic external FlashAttention is selected only when all of these hold:

1. SDPA was requested and legacy mode was not requested.
2. CUDA and both external FlashAttention entry points are available.
3. PyTorch reports no native fused FlashAttention backend.
4. The GPU is compute capability 8.0 or newer.
5. Fixed and variable-length BF16 GQA forward/backward probes succeed.
6. Actual runtime tensors satisfy FlashAttention's device, dtype, head-dimension, and grouping contract.

Any failed check retains the existing PyTorch SDPA path. Native fused SDPA platforms, unsupported GPUs, CPU execution, missing/broken FlashAttention installations, and explicit alternative attention modes keep their previous behavior.

## Validation

- Full clean-upstream test suite: `74 passed`
- Focused backend/parser/fallback tests: `14 passed`
- Ruff check on every changed Python file: passed
- Ruff format check on every changed Python file: passed
- Seven real no-block-swap legacy/automatic training pairs completed on Windows

Breaking changes: none.

### DIFF
--- a/src/musubi_tuner/hunyuan_model/attention.py
+++ b/src/musubi_tuner/hunyuan_model/attention.py
@@ -3,12 +3,14 @@ import math
 import torch
 import torch.nn.functional as F
 
+from musubi_tuner.modules.attention import AUTO_FLASH_ATTENTION_MODE, resolve_automatic_flash_mode
+
 try:
     import flash_attn
     from flash_attn.flash_attn_interface import _flash_attn_forward
     from flash_attn.flash_attn_interface import flash_attn_varlen_func
     from flash_attn.flash_attn_interface import flash_attn_func
-except ImportError:
+except Exception:
     flash_attn = None
     flash_attn_varlen_func = None
     _flash_attn_forward = None
@@ -127,6 +129,8 @@ def attention(
     q, k, v = q_or_qkv_list if type(q_or_qkv_list) == list else (q_or_qkv_list, k, v)
     if type(q_or_qkv_list) == list:
         q_or_qkv_list.clear()
+    if mode == AUTO_FLASH_ATTENTION_MODE:
+        mode = resolve_automatic_flash_mode(q, k, v)
     split_attn = total_len is not None
     if (split_attn or cu_seqlens_q is None) and mode == "sageattn":
         mode = "sageattn_fixlen"
--- a/src/musubi_tuner/hunyuan_model/models.py
+++ b/src/musubi_tuner/hunyuan_model/models.py
@@ -807,11 +807,11 @@ class HYVideoDiffusionTransformer(nn.Module):  # ModelMixin, ConfigMixin):
         max_seqlen_kv = max_seqlen_q
 
         attn_mask = total_len = None
-        if self.split_attn or self.attn_mode == "torch":
+        if self.split_attn or self.attn_mode in {"torch", "flash_auto"}:
             # calculate text length and total length
             text_len = text_mask.sum(dim=1)  #  (bs, )
             total_len = img_seq_len + text_len  # (bs, )
-        if self.attn_mode == "torch" and not self.split_attn:
+        if self.attn_mode in {"torch", "flash_auto"} and not self.split_attn:
             # initialize attention mask: bool tensor for sdpa, (b, 1, n, n)
             bs = img.shape[0]
             attn_mask = torch.zeros((bs, 1, max_seqlen_q, max_seqlen_q), dtype=torch.bool, device=text_mask.device)
--- a/src/musubi_tuner/hv_train.py
+++ b/src/musubi_tuner/hv_train.py
@@ -32,6 +32,7 @@ from diffusers.optimization import (
 from transformers.optimization import SchedulerType, TYPE_TO_SCHEDULER_FUNCTION
 
 from musubi_tuner.dataset import config_utils
+from musubi_tuner.modules.attention import resolve_sdpa_backend
 from musubi_tuner.modules.custom_offloading_utils import BlockSwapConfig
 from musubi_tuner.hunyuan_model.models import load_transformer, get_rotary_pos_embed_by_shape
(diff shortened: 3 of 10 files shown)


kohya-ss:
Thank you very much for this PR and for the careful implementation — the probe-guarded fallback and the test coverage are genuinely thoughtful work.

You are right about the underlying observation. On some Windows PyTorch builds, `--sdpa` does end up on the slower memory-efficient kernel because the native FlashAttention SDPA backend isn't present. I reproduced exactly this locally (`torch 2.12 + cu130`, `torch.backends.cuda.is_flash_attention_available() -> False` while CUDA is available), so the problem you're targeting is real.

After reviewing it end to end, though, I've decided not to merge this, and I'd like to explain why honestly:

- **It largely duplicates an existing option.** When the external `flash_attn` package is installed and works, users can already select it explicitly with `--flash_attn`. The core value this PR adds is auto-swapping `--sdpa` to flash behind the scenes, which is a convenience rather than a new capability.

- **It silently overrides an explicit user choice.** When someone passes `--sdpa`, they may have chosen it deliberately (determinism, matching the train/inference backend, or avoiding a flash-specific issue). FlashAttention and SDPA are not numerically identical, so silently switching backends changes training results, with only a somewhat hidden opt-out (`--use_legacy_sdpa` / env var). For a training framework I'd want the default to preserve the requested backend.

- **The wiring isn't complete across architectures, and this causes a regression.** `_load_dit_and_swap` is shared by all `NetworkTrainer` subclasses, so every architecture can now receive the new `"flash_auto"` mode, but only three attention dispatchers understand it (`modules/attention.py`, `hunyuan_model/attention.py`, `wan/modules/attention.py`). FramePack, for example, asserts on unknown modes (`frame_pack/hunyuan_video_packed.py`), so a Windows user with `flash_attn` installed running FramePack with `--sdpa` would crash where it previously worked. Kandinsky's separate engine has a similar risk.

- **The ongoing maintenance cost outweighs the benefit.** Introducing a global pseudo-mode means every current and future attention dispatcher has to keep understanding `"flash_auto"`, plus the per-device probe and the cross-trainer flag. That's a fair amount of cross-cutting surface for a convenience that, on the newest GPUs, often doesn't even apply (the external kernels may not support the architecture either, so the probe fails and it falls back to torch anyway).

Given all of that, the most honest fix for the actual pain point is probably documentation: noting that on Windows, `--flash_attn` should be preferred over `--sdpa` when the FlashAttention package is available. That solves the real case with zero blast radius.

If we did want automatic selection, I think the more valuable and cleaner design would be one level up: For example, a single `--auto_attention_mode` that genuinely picks the best *available* backend (torch/sdpa, flash, sageattn, …) and behaves exactly as if the user had specified it — rather than special-casing SDPA→flash with a new mode threaded through the shared core. That's something I'd rather design and own on our side so it stays consistent across all architectures.

I really appreciate the effort here and I don't want to discourage further contributions — this just isn't the right shape for the codebase. Thank you again!

FurkanGozukara:
sure thanks a lot for reply. it was really extra info for me . finally i know windows linux major difference reason :D

PR #1008, opened 2026-07-21
feat: ConvRot int8 quantization for Krea 2 LoRA training

kohya-ss:
## Overview

Adds `--convrot_int8`: ConvRot int8 quantization ([arXiv:2512.03673](https://arxiv.org/abs/2512.03673)) of the frozen DiT base weights for Krea 2 LoRA training, as an alternative to `--fp8_base --fp8_scaled`.

A block-diagonal Hadamard rotation (group size 256) smooths activation outliers before quantization. Offline, weights are rotated and quantized per-channel to int8; online, the forward pass runs a fused Triton kernel (activation rotation + dynamic row-wise int8 quantization + int8 GEMM with the dequantization epilogue fused in). The backward is a custom `autograd.Function` — the base is frozen, so only `grad_x = rotate(g @ W_rot)` is needed — with a `bf16` mode (default, transient dequant) and an opt-in `int8` mode (`--convrot_int8_bwd int8`, reuses the fused GEMM).

The main benefit is **speed on GPUs without fp8 support** (RTX 30 series and older), where fp8 can only dequantize to bf16 while ConvRot runs a true int8 GEMM. Weight VRAM halves vs bf16, same as fp8.

## Measured (LoRA training, 1024x1024, batch size 1, `--flash_attn`, gradient checkpointing)

| GPU | bf16 | fp8_scaled | convrot_int8 |
|---|---|---|---|
| RTX 3090 | (does not fit) | 7.1 s/step | 5.3 s/step |
| RTX PRO 6000 Blackwell Max-Q | 2.0 s/step | 2.3 s/step | 1.9 s/step (1.8 with `--compile`) |

Loss curves match the fp8/bf16 baselines closely, in both backward modes.

## Implementation notes

- `modules/convrot_int8_kernels.py`: Triton int8 kernels + Hadamard helpers vendored from [comfy-kitchen](https://github.com/Comfy-Org/comfy-kitchen) (Apache-2.0, headers preserved, noted in README License). triton is optional — without it, forward falls back to transient dequantized bf16 matmul (VRAM savings kept, no speedup).
- `modules/convrot_int8_utils.py`: streaming quantizer, autograd.Function, monkey patch. Follows the fp8 pattern — module stays `nn.Linear` with int8 `.weight` + `.scale_weight` buffer, so LoRA wrapping, block swap weight streaming, and compile exclusion keep working unchanged. The forward replicates `F.linear`'s autocast input casting (K2's fp32 modulation promotes activations to fp32; without the cast, flash-attn rejects the fp32 q/k/v and SDPA silently computes attention in fp32).
- `lora_utils.load_safetensors_with_lora_and_fp8` gains a generic `quantizer=` seam; LoRA merges into the bf16 weights before quantization; the fp8 path is untouched. The seam is reused for the planned pre-quantized-checkpoint loading and other architectures.
- `dit.requires_grad_(False)` before `load_state_dict(assign=True)`: int8 tensors cannot be wrapped as Parameters with `requires_grad=True` (fp8 never hit this because float8 is a floating dtype).
- Also fixes a latent issue in `load_transformer`: with `--blocks_to_swap`, quantization ran on the CPU (the calc device was set to the loading device). It now quantizes on `accelerator.device` and returns the weights to CPU. This also affected `--fp8_scaled`, where the simpler kernels made it less noticeable; the ConvRot rotation matmuls made it obvious.
- Not supported together with `--turbo_dit` yet (rejected with a clear error; needs the quantizer threaded through the Turbo/RAW stashes — small follow-up). Multi-GPU (DDP) is untested.

## Tests

`tests/test_krea2_convrot_int8.py` (20 tests): forward/grad_x parity vs the dequantized reference, LoRA A/B gradient parity, autocast casting (CPU + CUDA), eager fallback parity, quantization scope (excluded / indivisible layers), streaming-quantizer → monkey-patch → `load_state_dict(assign=True)` round-trip, non-reentrant gradient checkpointing, and trainer flag validation.

## Follow-ups (planned on dev)

- Phase 2: load pre-quantized ComfyUI ConvRot int8 checkpoints (`weight` int8 + `weight_scale` + `comfy_quant` metadata)
- Phase 3: roll out to other architectures
- `--turbo_dit` support

🤖 Generated with [Claude Code](https://claude.com/claude-code)


recris:
I'd love to try this one on AMD, the 7000 series (RDNA3) also have INT8 hardware instructions, and we're paying a performance penalty by using FP8 weights just to fit things in VRAM.

rockerBOO:
Trying this out. 

*Edit*: Removed issues I ran into because I was running it on a different branch by accident

```toml
flash_attn = true

gradient_checkpointing = true
batch_size = 1

blocks_to_swap = 26
block_swap_h2d_only = true
block_swap_ring_size = 2 # 1 compute, 1 prefetched
use_pinned_memory_for_block_swap = true

convrot_int8 = true
convrot_int8_bwd = "int8" # default bf16
```

Running at 11GB roughly for 1024 resolution, 8 rank, with 26 block swap on a 5070ti. 

```
 46/785 [02:33<40:58,  3.33s/it
```

As a side question, is this bypassing or ignoring the mixed precision argument to accelerate? One of the issues I found previously was the mixed precision was possibly upsampling the fp8 weights to bf16 because of the mixed precision option in accerate. On hardware that supports fp8 i'd probably want it to be fp8. Getting accelerate working with mixed precision fp8 was pretty annoying previously because you need third party dependencies. I can do a proper investigation at some point.  This int8 version seems to be going 2x faster than the fp8 version with mixed precision of bf16.

rockerBOO:
Screenshot 2026-07-31 at 03-52-31 summer-voice-37 Rockerboo&#39;s workspace retro-anime-krea-2-kohya-lora – Weights   Biases

Noticing a pattern right after sampling (at step 250) where the gradients are spiking. since it is right after sampling images are made, might be something to look into for something happening during that. I'm running `schedulefree.RAdamScheduleFree` optimizer though I do not think this is directly related.

Edit: Now realizing I'm also running torch compile on this so when it does the sampling it usually does a recompile. But this might be having some side effect too. I can look more deeply into this issue but just reporting for now. 

Edit 2: Second sample happened at 500 but didn't spike noticeably there though. Maybe just coincidence but just reporting again.
Screenshot 2026-07-31 at 04-29-41 summer-voice-37 Rockerboo&#39;s workspace retro-anime-krea-2-kohya-lora – Weights   Biases

Edit 3: I did an investigation and at this point it seems coincidental but will keep an eye on it. Since it only happened on the first sampling but the second didn't. And other runs didn't have this pattern either.

FurkanGozukara:
i  have implemented this into my app

results are amazing

more closer to BF16 than FP8 scaled

1.5x speed up

image

image


image

image

PR #1014, opened 2026-07-30
fix: cast non-target fp32 weights to dtype in ConvRot int8 quantizer

rockerBOO:
Ran into a dtype issue when running `--sample_at_first` and running the `--convrot_int8`. This PR looks to cast these appropriately. Maybe this could be "fixed" by setting things up before the sample images are done but this PR does fix the issue.

## Summary
- `ConvRotInt8Quantizer.load_and_quantize` left excluded/indivisible/non-target keys (e.g. K2's `first` patch-embed Linear) at whatever dtype the checkpoint stored, since `dit_weight_dtype` was never threaded through for the quantized load path (`load_krea2_dit` always passed `dit_weight_dtype=None` when `quantized` is true, regardless of `fp8_scaled` vs `convrot_int8`).
- When a checkpoint stores those non-target weights in fp32 (as the K2 checkpoint does), this breaks `F.linear` with a `BFloat16`/`Float32` dtype mismatch once the rest of the model runs bf16 — hit this training a K2 LoRA with `--convrot_int8` enabled, failing at step-0 sample-image generation inside a LoRA-wrapped `first` Linear.
- `ConvRotInt8Quantizer` now accepts a `dtype` and casts any non-quantized floating weight to it in both non-target and indivisible-weight branches; `krea2_utils.load_krea2_dit` passes the real `dtype` through for `--convrot_int8`. The `--fp8_scaled` path is untouched — its checkpoint-dtype behavior there is documented as intentional in the existing docstrings/comments.

## Test plan
- [x] Added `test_quantizer_casts_non_target_fp32_weights_to_dtype`, reproducing a checkpoint with fp32 non-target/indivisible weights and asserting they're cast when a `dtype` is supplied; fails on `main`/pre-fix code (`TypeError: unexpected keyword argument 'dtype'`), passes after the fix.
- [x] `uv run --no-sync --with pytest pytest tests/test_krea2_convrot_int8.py -q` — 21 passed (all existing ConvRot int8 tests plus the new one).

### DIFF
--- a/src/musubi_tuner/krea2/krea2_utils.py
+++ b/src/musubi_tuner/krea2/krea2_utils.py
@@ -113,7 +113,7 @@ def load_krea2_dit(
             target_keys=KREA2_FP8_OPTIMIZATION_TARGET_KEYS if fp8_scaled else None,
             exclude_keys=KREA2_FP8_OPTIMIZATION_EXCLUDE_KEYS if fp8_scaled else None,
             quantizer=(
-                ConvRotInt8Quantizer(KREA2_FP8_OPTIMIZATION_TARGET_KEYS, KREA2_FP8_OPTIMIZATION_EXCLUDE_KEYS)
+                ConvRotInt8Quantizer(KREA2_FP8_OPTIMIZATION_TARGET_KEYS, KREA2_FP8_OPTIMIZATION_EXCLUDE_KEYS, dtype=dtype)
                 if convrot_int8
                 else None
             ),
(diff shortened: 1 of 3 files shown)


rockerBOO:
Closing this because by setting mixed-precision to "no" in accelerate basically caused this and we'd still need to do some conversion between the LoRA weights (fp32) and the int8 if we don't use the mixed precision to autocast them to be the same. by having accelerate handle it we could support fp8 as a mixed-precision so making this possibly more flexible.

PR #1018, opened 2026-08-03
Add MiniMax-H3 BF16 training and inference support

sdbds:
## Summary

This Draft PR implements native **BF16 MiniMax-H3 R1** support in Musubi Tuner.

R1 includes:

- T2VA, FL2VA, and JSONL-only Ref2VA latent caching
- Qwen3-VL-32B multimodal text caching with per-row modality tags
- BF16-base LoRA training with joint video/audio flow loss
- standalone and scheduled training-time joint video/audio generation with PyAV muxing
- published Comfy-Org BF16 FL2VA and Ref2VA transformers plus the released video/audio VAEs
- classic and H2D-only block swap for the 50 H3 main blocks
- strict dataset `batch_size = 1`; gradient accumulation is the R1 effective-batch mechanism

Real packed multi-sample batching is deferred to a separate PR. Structural interfaces still retain explicit leading batch axes for positions, token tags, and row timestep indices.

## Implementation

- Registers `mmh3` / `minimax_h3`, 24 fps, 32-pixel buckets, and architecture-aware `17*n+5` frame normalization.
- Uses the existing cache/collator contracts: target video becomes `batch["latents"]`, audio remains released `[32,2,A]`, and variable text rows use `varlen_mmh3_*`.
- Supports embedded target audio, explicit JSONL audio, and one same-stem sidecar; Ref2VA preserves ordered image/video/audio references.
- Captures the unnormalized Qwen state after layer 50 (`hidden_states[50]`) and enforces a 32768-row limit.
- Reproduces the released packing, rotary clock, three-modality block AdaLN, one-modality FinalLayer AdaLN, and separate video/audio row times.
- Trains raw dataward heads against `latents - noise` with unweighted `mean(video_mse) + mean(audio_mse)`.
- Samples on one common base grid with independent shifted video/audio sigma intervals. It does not copy ComfyUI's output negation or single-sampler audio slope adapter.
- Standalone generation loads Qwen3-VL and both VAEs sequentially around the 33B transformer. Training-time sampling prepares Qwen3-VL states and condition latents before transformer allocation, retains both VAEs on CPU, and moves the decode VAEs to the accelerator one at a time.
- Validates generation text caches by task, frame count, structural metadata, and exact presentation fingerprint.
- Overrides the shared single-VAE/video-only sampling hooks: scheduled samples use the live training transformer and LoRA, jointly denoise video/audio latents, decode both modalities sequentially, and mux a voiced MP4 while retaining the shared cadence, distributed assignment, RNG restoration, and block-swap transitions.

## Provenance

The transformer, video VAE, packing, sampling, and relevant text-encoder logic are adapted from Apache-2.0 [Diffusers PR #14355](https://github.com/huggingface/diffusers/pull/14355), pinned at commit `abc5e9bf71fd38f53cd471bc3acaa84bc5ecbfdc`. The modules retain upstream copyright/license headers and map their source files in the header.

ComfyUI is used only as an independent numerical and artifact-compatibility reference. Its GPL-3.0 implementation is not a code source. The video VAE implementation was rewritten from the Diffusers source while preserving the published checkpoint state-dict schema.

## Documentation

- User guide: `docs/minimax_h3.md`
- Reviewed design: `docs/superpowers/specs/2026-08-03-minimax-h3-support-design.md`
- Implementation plan/evidence: `docs/superpowers/plans/2026-08-03-minimax-h3-r1-implementation.md`

## Verification

- `python -m pytest -q`: **222 passed**
- Ruff 0.15.12 format check: **278 files already formatted**
- Ruff 0.15.12 check: **all checks passed**
- Rewritten video VAE versus the pre-rewrite artifact-matched definition: **562/562 state keys with identical shapes and dtypes**; tiny encoder/decoder numerical parity passed
- Generation entrypoint `--help`: passed
- Real PyAV mux smoke: produced H.264 video and AAC audio streams
- Published safetensors header parity against the model definitions:
  - FL2VA BF16 transformer: 535 tensors, 0 missing/unexpected/shape/dtype mismatches
  - Ref2VA BF16 transformer: 535 tensors, 0 missing/unexpected/shape/dtype mismatches
  - video VAE: 562 tensors, 0 missing/unexpected/shape mismatches
  - audio VAE: 917 tensors, 0 missing/unexpected/shape mismatches
  - Qwen3-VL-32B BF16: 902 tensors, 0 missing/unexpected/shape mismatches

The complete published BF16 artifact set is approximately 120 GB and is not present on the validation machine, so a tensor-body real-model forward/sample has **not** been claimed. The PR remains open pending maintainer review and that hardware/artifact smoke.

## Deferred R2

ConvRot, prequantized INT8, runtime LoRA over prequantized weights, pruned AdaLN, FP8, and NVFP4/AWQ are deferred until #1008 is merged upstream and its final interfaces can be reviewed. R1 has no dependency on the unmerged PR.

## References

- Published artifacts: https://huggingface.co/Comfy-Org/MiniMax-H3
- Original model: https://huggingface.co/MiniMaxAI/MiniMax-H3
- Apache implementation lineage: https://github.com/huggingface/diffusers/pull/14355
- Numerical reference only: https://github.com/Comfy-Org/ComfyUI/pull/15224
- Deferred ConvRot foundation: #1008


sdbds:
@kohya-ss Could you please confirm whether this design direction is acceptable before implementation starts?

The main scope decisions I would like to confirm are:

1. Native Musubi support for T2VA, FL2VA, and Ref2VA, including latent/text caching, LoRA training, and joint video/audio inference.
2. LoRA-only training in the first release; no full transformer fine-tuning.
3. Merge/reuse #1008 for the ConvRot foundation, then add MiniMax-H3 support for the published normal and pruned prequantized ConvRot checkpoints.
4. Transformer block swap supported in both training and inference.
5. Do not impose a MiniMax-H3-specific `batch_size == 1` restriction; retain only normal compatible-bucket batching.

The design spec in this Draft PR contains the file layout, data/reference schema, checkpoint matrix, packing/loss contract, tests, and explicit non-goals. I will adjust or split the implementation if you prefer a smaller upstream scope.

sdbds:
@kohya-ss The RFC has been revised after a repository-contract review and the proposed first release is now narrower:

- R1 is BF16 only: T2VA / FL2VA / JSONL-only Ref2VA, LoRA training, joint video/audio inference, and block swap.
- ConvRot, prequantized INT8, runtime LoRA over INT8, and pruned AdaLN are deferred until #1008 is merged and its final API is available.
- The cache keys now follow the existing `latents_` / `varlen_` collator contracts exactly.
- H3 explicitly owns audio-noise creation and overrides both dual-modality batch processing and unweighted loss reduction.
- No layout-signature bucket, directory reference parser, text-padding system, or forced `batch_size == 1` check is proposed.

Could you confirm whether this BF16-only R1 is an acceptable upstream scope before the implementation plan is added?

sdbds:
@kohya-ss One more material spec update after auditing the now-merged ComfyUI MiniMax-H3 implementation (ComfyUI#15224, merge `57500fc5`):

- raw H3 heads and LoRA targets are explicitly `latents - noise`; ComfyUI's output negation and audio `d sigma_a / d sigma_v` slope are documented as single-sampler adapters and are not copied into native Musubi training/inference
- condition augmentation, `max(model_t, aug)` row times, per-token mixed text/vision tags, and the exact rotary clock are now pinned with golden-test requirements
- video/audio shifts are configurable; native inference uses two finite sigma schedules and is not treated as bit-exact with ComfyUI's single-scheduler trajectory
- there is still no forced `batch_size == 1` check, but compatible multi-sample forwards share one layout and one base timestep so no per-sample timestep is silently dropped

The proposed upstream scope remains BF16-only R1 with LoRA, joint AV inference, and block swap; ConvRot/pruned work remains deferred behind #1008. The updated design is in `af08a0d`. Please confirm whether this R1 boundary is acceptable before implementation begins.

kohya-ss:
Thank you for this PR! I will check this and comment as soon as I have time.

FurkanGozukara:
awesome looking forward this to mature

LTX 2.3 training working great atm

sdbds:
@kohya-ss The BF16 R1 implementation is now pushed and the Draft PR body has been updated for review. Automated verification is 212 passed with Ruff clean. I also checked the published Comfy-Org safetensors headers against the actual model definitions: both BF16 transformers, both VAEs, and the BF16 Qwen3-VL file have zero key/shape mismatches (and zero transformer dtype mismatches). The full ~120 GB artifact set is not local, so I have explicitly left the real tensor-body forward/sample smoke open rather than claiming it. ConvRot/pruned support remains deferred behind #1008.

kohya-ss:
Thank you for the detailed RFC and for the careful audits — especially the separation of ComfyUI's sampler-side adapters (output negation, audio `dσ_a/dσ_v` slope) from the model's native `latents - noise` semantics, and the R1/R2 split. I can now confirm the scope.

**Confirmed scope**

1. Native T2VA / FL2VA / JSONL-only Ref2VA support with latent/text caching, LoRA training, and joint AV inference: **agreed**.
2. LoRA-only training in the first release: **agreed**.
3. ConvRot / prequantized INT8 / pruned AdaLN deferred to R2 until #1008 is merged, with R1 having no dependency on it: **agreed**. This is the right call — #1008's final API may still change.

**Requested change 1: simplify batching to `batch_size == 1` in R1**

I'd like to replace the "replicated batch axis + post-build preflight" design (spec §8.3 / §12.2 / §15, `validate_h3_dataset_batches`, and the related tests) with a simple check that H3 datasets use `batch_size = 1`, with gradient accumulation as the batching story for R1.

Rationale: since a multi-item batch requires identical text length *and* an identical token-tag vector, real-world samples with different captions will almost never batch together. The preflight machinery is significant complexity (safetensors header scans over every multi-item bucket, timestep-pool rejection, many tests) in service of a case that effectively only occurs with duplicated items.

Real batching for t2va/fl2va (text padding + attention mask inside the packed sequence, per-sample structural tensors) is worth doing later, but as a separate PR. To keep that door open cheaply, please keep the packed structural tensors (`position_ids`, `token_tags`, `row_timestep_indices`) shaped/passed per-batch (`[B, S]`-style interfaces) even while R1 only ever builds one shared plan. Ref2VA will likely stay batch-1 regardless, since heterogeneous reference layouts can't be absorbed by padding alone.

**Requested change 2 (important): code provenance and licensing**

The implementation plan states the transformer and VAEs are ported from the merged ComfyUI implementation. ComfyUI is GPL-3.0, while this repository is Apache-2.0, so a direct port/translation of ComfyUI code cannot be merged here.

Please base the ported modules (`model.py`, `video_vae.py`, `packing.py`, `sampling.py`, and the relevant parts of `text_encoder.py`) on Apache-2.0-compatible sources — the official MiniMaxAI repository and/or the Diffusers PR — and use ComfyUI only as a behavioral reference for numerical verification. `audio_vae.py` already documents its lineage (DAC/BigVGAN/alias-free, all MIT/Apache) — please add the same kind of provenance/license header to the other ported modules, and update the plan/spec wording accordingly. Small leftovers like `ops = nn` and the `LTXVAudioVAEDecode` comment also suggest a Comfy-derived starting point; please clean these up as part of the same pass.

**Notes (no action needed now)**

- `docs/superpowers/` can stay in the PR for now — it has been useful for review. I plan to remove it in a follow-up after merge, since `docs/` here is user-facing documentation; no need to relocate it yourself. User-facing docs per the existing `docs/` conventions are welcome.
- Shared-code touches (the architecture branch in `round_down_frame_count`, the H3 validation block in `cache_io.py`, the H3-specific text-cache filename branch in `image_video_dataset.py`) are acceptable for now; I'll adjust them toward this repo's extension patterns in a follow-up after merge, as usual.
- With `frame_extraction="full"`, an H3 video shorter than 5 frames currently raises and aborts the whole caching run; skipping with a warning would be friendlier, but this can also be a follow-up.
- Restricting `--timestep_sampling` to `uniform` and `--weighting_scheme` to `none` in R1 is fine with me.
- The strict checkpoint loader assumes the released artifacts contain all persistent buffers (`latents_mean/std`, `mask_token`, `zero_k_bias`, ...); the real-artifact validation in Task 10 will be the actual gate here, so please include those results in the PR when ready.

With the two requested changes above, please go ahead — I'm happy with the overall direction and the implementation quality so far (I also ran the full test suite at the current head: 210 passed).

sdbds:
Final review hardening is in b658f0c.

- Effective batch size 1 now reads only authoritative latent task metadata; replicated buckets alone build full fingerprints, with token tags reduced to SHA-256 and no dataset-wide fingerprint cache.
- AdaLN run discovery is vectorized; layout/device/dtype rotary tables use a bounded two-entry cache that is cleared on moves and checkpoint loads.
- rope.inv_freq is checkpoint-owned (torch.empty), gate modulation no longer concatenates every segment, and standalone training/sampling tests set up src themselves.
- Equal modality loss and the two intentional non-bitwise ComfyUI differences are now explicit in metadata/docs.

Two reported items were verified rather than changed: t2va intentionally has task=t2va with base_family=fl2va because it uses the released FL2VA base; trainable AdaLN scale/shift/gate backward matches the out-of-place reference, so the claimed in-place autograd failure was not reproducible. Complete timestep rows remain dynamic because both model times change each step; only truly invariant rotary state is cached.

Verification: 220 tests passed; Ruff 0.15.12 check . passed; the touched review files also pass Ruff 0.16.1. The existing workflow still installs unbounded latest Ruff and therefore reports unrelated repository-wide 0.16 rule expansion; pinning that tool version remains a separate CI decision.

sdbds:
@kohya-ss Both requested changes are implemented in `c4dc6b8`.

**R1 batching**
- Every H3 dataset now requires configured `batch_size = 1`; the error points to gradient accumulation.
- `validate_h3_dataset_batches`, all safetensors fingerprint/header scanning for training compatibility, the replicated-forward path, and its positive multi-item tests were removed.
- Runtime, sampler, and model keep defense-in-depth `B == 1` checks.
- `position_ids` remain `[1,S,3]`; token tags, row timestep indices, and AdaLN modulation plans retain explicit leading batch structure so a later real-batching PR does not inherit unbatched interfaces.

**Provenance**
- `model.py`, `video_vae.py`, `packing.py`, `sampling.py`, and the relevant text path now identify Apache-2.0 Diffusers PR #14355 at pinned commit `abc5e9bf71fd38f53cd471bc3acaa84bc5ecbfdc`, with per-file source mappings and retained license headers.
- ComfyUI is documented and tested as numerical/artifact validation only.
- The video VAE was substantively rewritten from the Diffusers implementation while preserving all 562 published state keys, shapes, and dtypes; tiny encoder/decoder parity checks pass.
- Removed the `ops = nn` alias and the unrelated `LTXVAudioVAEDecode` name.

Verification at the pushed head: 218 tests passed; Ruff 0.15.12 check/format passed; GitHub Ruff CI passed. The PR body and design/plan/user docs now reflect the smaller R1. The full real-weight forward/sample remains explicitly open.

sdbds:
@kohya-ss Follow-up 548275a restores the training-time sampling path that was present in the approved design but was accidentally omitted from the implementation plan and then rejected in ca801ba.

- H3 now overrides the shared single-VAE/video-only preparation and per-prompt inference hooks instead of rejecting --sample_prompts.
- Qwen3-VL states and FL2VA/Ref2VA condition latents are prepared before transformer allocation; the video/audio VAEs remain on CPU and move to the accelerator sequentially for decode.
- Scheduled samples use the live transformer and current LoRA, jointly denoise video/audio, mux a voiced MP4, and retain the shared cadence, distributed assignment, RNG restoration, and block-swap mode transitions.
- T2VA, FL2VA, and JSONL Ref2VA prompt contracts are documented; CFG/negative prompts remain unsupported.

Verification at the pushed head: 222 tests passed; all 149 H3 tests passed; Ruff format/check passed. The PR body has been updated.

kohya-ss:
Thank you for the quick turnaround — I've reviewed all four commits since my last comment.

**Both requested changes are confirmed as addressed in `c4dc6b8`:**

1. **Batching**: the preflight machinery is fully removed, the `batch_size = 1` validation with the gradient-accumulation guidance is in place, and the structural tensors keep their explicit `[1, S]` batch axes as discussed. The defense-in-depth `B == 1` checks in the runtime/sampler/model are a good balance.
2. **Provenance**: I verified the Diffusers PR #14355 lineage and the pinned commit, and that `video_vae.py` was genuinely restructured rather than relabeled. The `ops = nn` / `LTXVAudioVAEDecode` leftovers are gone, and the provenance regression test is a nice touch. This resolves the licensing concern.

I also went through `b658f0c` (the bounded rotary cache with invalidation on `_apply`/`load_state_dict` looks correct) and `548275a`. The training-time sampling restoration is well designed — keeping the base `sample_images` ownership of RNG save/restore and the block-swap transition, preparing text states before the transformer is allocated, and extracting the shared helpers into `generation_inputs.py` instead of duplicating them. The Ruff `<0.16` pin is fine with me. On my side the full suite passes at the current head (222 passed) and Ruff check/format are clean.

**Task 10 (real-artifact validation): I'll run it on my side.**

Since the full artifact set is not available to you locally, I've downloaded the released checkpoints (both BF16 transformers, the Qwen3-VL text encoder, and both VAEs) and will run the smoke here, then post the results in this PR. Planned coverage, starting with t2va:

1. `minimax_h3_cache_latents.py` on a small real video with embedded audio (also exercises the strict loader's buffer assumptions)
2. `minimax_h3_cache_text_encoder_outputs.py` with the real Qwen3-VL weights
3. A short LoRA training run with block swap (real forward/backward, loss sanity)
4. One scheduled training-time sample through the new `548275a` path (sampling → both VAE decodes → muxed MP4)
5. Ref2VA afterwards with the second transformer

One item I'd like to fold into the same run: **video VAE dtype**. This PR runs the video VAE in fp16 across all paths, while the Diffusers PR the port is now based on pins both VAEs to fp32 (see its "Keep VAEs in float32" commit). Was fp16 a deliberate choice (e.g., following ComfyUI's runtime behavior)? I'll compare fp16 vs fp32 decode numerically as part of the smoke either way — if fp16 shows instability on real tensors we can switch the default, and if it's clean we'll document the deviation.

One note for later: my test machine has ample RAM, so this smoke won't tell us how the BF16-only R1 behaves on the ≤64 GB systems most users have. I'll treat that as a post-merge follow-up rather than a gate for this PR.

kohya-ss:
**Task 10 real-artifact smoke: first results (t2va complete)**

Environment: 96 GB VRAM GPU, 512 GB RAM, Windows. Test data: a synthetic 384x224, 39-frame (1.625 s) clip with embedded stereo 32 kHz audio, `--task t2va`, LoRA dim 16, 60 steps. Small geometry via `--allow_experimental_duration`.

The full pipeline now runs end-to-end with the released checkpoints: latent caching → text-encoder caching → LoRA training (all three block-swap configurations) → training-time joint AV sampling → final save → standalone generation with the trained LoRA. Getting there surfaced four bugs, all shallow and none affecting the model math:

1. **`minimax_h3_train_network.py` has no `if __name__ == "__main__": main()` block**, so the documented `accelerate launch` command exits silently with code 0. The other three entry points have the guard. (The test suite imports the module and calls functions directly, which is why 222 passing tests never noticed — same for the items below.)
2. **`install_h3_video_decoder`'s patched `get_video_data_from_path` drops the `self.bucket_selector` fallback** that the base datasource provides (datasources.py:431). Real directory-dataset caching always reaches it with `bucket_selector=None` and crashes with `AttributeError: 'NoneType' object has no attribute 'get_bucket_resolution'`. Restoring the fallback (`if bucket_selector is None: bucket_selector = datasource.bucket_selector`) fixes caching.
3. **`utils/sai_model_spec.py` doesn't register `mmh3`**, so training completes and then crashes in `build_metadata` at the final save — the trained LoRA is lost. Needs `ARCH_MINIMAX_H3` / `IMPL_MINIMAX_H3` constants and an `elif` branch like the other architectures.
4. **Docs: both TOML examples in `docs/minimax_h3.md` use `source_fps = 24`** — the schema requires a float, so copying the example fails validation. Please change to `24.0`.

I've applied 1–3 locally to keep the smoke moving; please push the proper fixes to this branch (happy to share my exact patches if useful).

**Results, for the record**

- Latent + TE caching: OK with real VAEs / Qwen3-VL. TE encode of one short caption: 14 s on CPU, ~64 GB RAM.
- Training: loss behaves (typically <1 with occasional high-timestep spikes, as expected for uniform sampling). Speeds at this geometry: classic swap 48 = 27 s/it, `--block_swap_h2d_only` + swap 32 = 8 s/it, no swap = 1.5 s/it at 65 GB VRAM.
- Training-time sampling works in both swap modes and produces real video+audio (not black/silent); sample content visibly changes by step 50.
- Saved LoRA: 600 tensors, exactly the four documented target modules, metadata correct (after fix 3).
- Standalone generation with and without the LoRA: OK at 2.24 step/s. The LoRA output is measurably closer to the training-time sample than the no-LoRA baseline (mean |diff| 51 vs 75/255), so the merge path and the training-time sampling path agree directionally.
- **VRAM note**: training-time sampling loads the 48 GB Qwen3-VL onto `accelerator.device`, which restricts `--sample_prompts` to ~50 GB+ GPUs. A device override (like the TE cache script's `--device`) or reuse of the cached text-encoder outputs would fix this — fine as a follow-up, but worth a note in the docs for R1.

**fp16/fp32 video VAE comparison** (the dtype question above)

- **Decode**: fp16 is safe — max 2.5/255, mean 0.04/255 vs fp32 (PSNR 73.6 dB), no NaN/Inf. Visually indistinguishable even at 20x amplification.
- **Encode** (deterministic posterior mean): mostly close (mean |diff| 0.005 in normalized units), but with a heavy tail — 0.32% of elements deviate by >0.1, eleven by >0.5, max 1.07 (latent std ≈ 1.32), concentrated in a few channels and including sign flips. Since cached latents are the training target, I'd prefer **fp32 for encode/caching** (one-time cost, VAE is small) while keeping fp16 decode — unless the fp16 behavior is itself the released semantics you pinned deliberately (I noticed `encode_video_condition`'s explicit fp16 roundtrip); please confirm.

Ref2VA smoke with the second transformer is next; I'll follow up with those results.

rockerBOO:
Is this being tested for only video input or is it suppose to be working with video and audio at this stage?

sdbds:
> Is this being tested for only video input or is it suppose to be working with video and audio at this stage?

Audio input is supported in the R1 target, typically using the video's built-in audio simultaneously. However, it seems this project didn't include audiodatasets, which might be an issue.

sdbds:
@kohya-ss Thank you for the real-artifact smoke and the precise repros. I have addressed all four failures in `df52b16`:

- added the missing direct-execution guard to `minimax_h3_train_network.py`
- restored the datasource `bucket_selector` fallback in the H3 decoder patch
- registered `mmh3` in SAI ModelSpec metadata
- changed both documented TOML values to `source_fps = 24.0`

I also adopted the hybrid video-VAE policy from your measurements: target and condition encoding now run in FP32, while decoding remains FP16.

I checked the pinned Diffusers implementation as well. Its keyframe path deliberately rounds the sampled condition latent through FP16 before normalization, but the VAE itself is kept in FP32. Therefore the explicit round-trip is released conditioning semantics, not a requirement to run the encoder weights in FP16. That round-trip remains unchanged.

The latent-cache `posterior_policy` now records `video_vae=fp32`, so caches produced by the former FP16 encoder path are invalidated under `--skip_existing` and should be rebuilt.

The R1 documentation now also calls out the roughly 50 GB accelerator-memory requirement of training-time sample prompt encoding.

Verification: 227 tests passed; Ruff check and format check are clean.

@rockerBOO Clarification on my earlier note: by "no audio datasets" I meant that Musubi's dataset abstraction currently exposes image/video datasets but no standalone audio dataset. R1 is not video-only: target audio aligned to each video is read from an embedded stream, JSONL `audio_path`, or a same-stem sidecar, and video/audio are trained and generated jointly.

Missing target audio still remains an error in R1. A follow-up goal is to accept silent source videos rather than reject them; whether that should synthesize supervised silence or mask the audio loss for video-only training needs a separate design decision and is deliberately not decided in this fix.

kohya-ss:
**Review of the seven commits since my smoke report — all confirmed**

I went through `df52b16` through `c4e040b3`:

- **`df52b16`**: all four reported failures are fixed exactly as expected (I diffed against my local workaround patches — identical semantics), and the hybrid VAE policy is cleanly applied via the `VIDEO_VAE_ENCODE_DTYPE`/`VIDEO_VAE_DECODE_DTYPE` constants across caching, generation, and training-time sampling. Your reading of the Diffusers keyframe path matches mine: the FP16 round-trip is conditioning semantics, the encoder weights themselves belong in FP32. The metadata-comparing `--skip_existing` is a nice upgrade over plain existence checks.
- **`4b0a016`**: the B=1 simplification of the modulation segments, the `load_generation_record` / `decode_generation_visuals` split (no more retaining raw pixels across model teardown), and the explicit `reference_video_frame_counts` contract all look right.
- **`c4e040b`** (video-only training): good design. Encoding real silence through the Audio VAE rather than assuming a zero latent, the per-sample binary weight with save/load-side consistency validation, the `supervised_audio_fraction` accounting, and the honest docs warning that a video-only LoRA leaves audio unconstrained (single-stream shared weights) — this is the right shape for R1, and the follow-up question about silent-video *supervision* stays open without blocking anything.

On my side at head `c4e040b3`: 257 tests passed, Ruff check and format clean.

**Ref2VA real-artifact smoke: complete — the full R1 matrix now passes end-to-end**

Setup: second (Ref2VA) BF16 transformer, synthetic 3 s target with embedded stereo audio, 39-frame crop via `--allow_experimental_duration`, and a reference set covering all three types in one record: a 256x256 image, a 2.528 s video with embedded audio (exercises the 2–15 s reference gate and the 60→39 frame trim), and a 2 s standalone WAV (audio-bearing count 2 of 3).

- Latent caching: OK, ~2.2 s for the record; all reference latents cached.
- TE caching: OK. The presentation now carries reference visuals — `rows=285, vision_rows=238`, 2.8 MiB payload, and ~4 min per record on CPU (vs ~14 s caption-only for t2va). See the follow-up note below.
- Training: 60-step LoRA on the Ref2VA base, no block swap: 4.1 s/it at ~70 GB VRAM / ~32 GB RAM. `supervised_audio_fraction=1.000000` logged at startup as expected. Training-time sampling at step 50 (record caption driving conditioning, no prompt override): ~1.2 s/step, produces real video+audio.
- Standalone generation, baseline vs trained LoRA (~1 s/step): the LoRA clearly changes the output, and the step-50 training-time sample sits closest to the training target — directionally consistent with the step-60 standalone LoRA output, same as the t2va result.

**Regression checks folded into this pass**

- Cache invalidation: re-running the old t2va caching with `--skip_existing` correctly logged `Rebuilding stale MiniMax-H3 latent cache` for the fp16-era cache (posterior_policy mismatch), and a second run logged `Skipping matching`. The mechanism works as advertised.
- Video-only path: `--h3_video_only` caching into a fresh cache directory reported `video_only=1` with `supervised_audio_fraction=0.000000`, the trainer logged the same fraction at startup, `loss/audio` stayed exactly 0 for every step (verified in the training logs), and a 10-step run completed normally.

**Follow-up (not a gate): quantized text-encoder artifacts**

The BF16-only Qwen3-VL is the practical bottleneck now: ~4 min per Ref2VA record on CPU, and the ~50 GB accelerator requirement for training-time sampling. ComfyUI ships INT8 ConvRot and NVFP4 TE weights for H3; supporting those for TE caching and sampling would make both usable on common GPUs. That was already R2 territory in the design doc — I just want to flag that Ref2VA caching cost makes it a higher-priority follow-up than I originally assumed.

With this, Task 10 is complete: both transformers, all three tasks' machinery, both cache-invalidation paths, and the new video-only mode have all been exercised against the released checkpoints. I'll do a final pass over the PR as a whole and then merge it into `dev`.

On R2 planning: the quantized-artifact work (INT8 ConvRot transformers, quantized TE) overlaps directly with the ConvRot INT8 infrastructure that is about to land in `dev` and with its planned rollout across architectures — and iterating on artifact loaders realistically requires having the released weights themselves on hand, which I know has been the hard part on your side. So I'd like to drive the R2 quantization work on my side as follow-ups on top of that infrastructure, keeping the supported formats to what users can actually obtain (INT8 ConvRot and NVFP4). Contributions on well-isolated pieces remain very welcome — pruned AdaLN in particular is largely structural and testable without the full artifact set, if you're interested. R1 as delivered here stands on its own; thank you for the quality and the turnaround throughout this PR.

rockerBOO:
The pruning for the AdaLN layers was [replaced by a lookup table](https://blog.comfy.org/i/209313677/optimized-for-local-inference-in-comfyui) so you'd need to glean that from the ComfyUI implementation for that to work. Since we are working on a 0 to 1 or 1 to 1000 timestep scale instead of a full continuous is why we can do this  in this case. Reduces memory requirements significantly though. I had an agent investigate it already for my quantization passes, can see it on <https://huggingface.co/rockerBOO/minimax-h3-nvfp4#pruned-variant-minimax_h3_fl2va_pruned_nvfp4safetensors>

Ada123-a:
> The pruning for the AdaLN layers

Here, take what you want from https://github.com/Ada123-a/musubi-tuner-H3_pruned_int8_prequantTE_image_training
It was a fork of AkaneTendo25's that I added pruned int8 checkpoint support to alongside image training.

Edit: AkaneTendo25 added all the important bits from it, use his fork now for reference instead https://github.com/AkaneTendo25/musubi-tuner/issues/106

MisticRain:
I tried training with AI-toolkit which forces using the pruned checkpoint to train on and on a 1270 video data set by epoch 2 characters started disintegrating I think training on the pruned int8 checkpoint is not a good idea. Will need to give this a go.

kohya-ss:
> The pruning for the AdaLN layers was [replaced by a lookup table](https://blog.comfy.org/i/209313677/optimized-for-local-inference-in-comfyui) so you'd need to glean that from the ComfyUI implementation for that to work. Since we are working on a 0 to 1 or 1 to 1000 timestep scale instead of a full continuous is why we can do this in this case. Reduces memory requirements significantly though. I had an agent investigate it already for my quantization passes, can see it on https://huggingface.co/rockerBOO/minimax-h3-nvfp4#pruned-variant-minimax_h3_fl2va_pruned_nvfp4safetensors

Thank you for letting me know! This is very helpful!

kohya-ss:
> > The pruning for the AdaLN layers
> 
> Here, take what you want from https://github.com/Ada123-a/musubi-tuner-H3_pruned_int8_prequantTE_image_training It was a fork of AkaneTendo25's that I added pruned int8 checkpoint support to alongside image training.
> 
> Edit: AkaneTendo25 added all the important bits from it, use his fork now for reference instead [AkaneTendo25#106](https://github.com/AkaneTendo25/musubi-tuner/issues/106)

Thank you. A lot of implementations has been done there. I will certainly use it as a reference.

Ada123-a:
> I tried training with AI-toolkit which forces using the pruned checkpoint to train on and on a 1270 video data set by epoch 2 characters started disintegrating I think training on the pruned int8 checkpoint is not a good idea. Will need to give this a go.

That is just due to the distillation of the model. There is a work around now, both in ai toolkit and musubi tuner. It's "H3 Guidance Distillation Scale" in AkaneTendo25's musubi fork. Set it to 3. I'm also testing using a loss from referencing the model with the lora disabled for about 5-10% of steps. Seems to completely fix the distillation issue at least up to a few thousand steps as I have not gone much beyond test character loras atm.

sdbds:
My Codex subscription just expired, and I plan to resubscribe after it resets. I think I've finished R1, and the rest is up to kohya and other contributors.
I also noticed that H3 can use SolAttn, and I think we can try replacing the position of sageattn for inference later.
https://github.com/kijai/ComfyUI-SolAttn_triton

rockerBOO:
I can donate some tokens if anything is scoped out. But is it basically testable at this point? Just bf16 is the only supported precision in this version?

sdbds:
> I can donate some tokens if anything is scoped out. But is it basically testable at this point? Just bf16 is the only supported version precision in this version?

yes, we need @kohya_ss  merge #1008 then test int8 convrot

kohya-ss:
Thank you, everyone.
I plan to merge this PR with #1008 by tomorrow morning, and if there are no issues, I should be able to complete all R2 work by tomorrow.

PR #1024, opened 2026-08-08
Add MiniMax-H3 INT8 ConvRot support

sdbds:
## Summary

- add automatic ConvRot artifact inspection and streamed loading based on the format introduced for Krea2 in #1008
- support the released full and pruned MiniMax-H3 FL2VA/Ref2VA INT8 ConvRot transformers
- support the released Qwen3-VL-32B text encoder INT8 ConvRot artifact
- integrate ConvRot bases with LoRA training, standalone generation, gradient checkpointing, compile exclusions, and CUDA block swap
- preserve FP32 scales and fixed-precision islands while converting only eligible F16/BF16 compute parameters

ConvRot is detected from the safetensors structure and each Linear's `.comfy_quant` control payload. No format-enabling flag is required. Malformed, partial, or unsupported quantized artifacts fail loudly instead of falling back to another interpretation.

## Pruned transformer support

Pruned artifacts are classified structurally and use the released FP32 `[1025, 8]` AdaLN curve table. Interpolation is performed in FP32, then transferred to the execution device and cast to the model compute dtype before AdaLN projection.

The loader validates the exact released topologies:

- full FL2VA/Ref2VA transformer: 250 ConvRot Linear layers
- pruned FL2VA/Ref2VA transformer: 200 ConvRot Linear layers
- Qwen3-VL text encoder: 350 ConvRot Linear layers

## Official artifact validation

HTTP range requests were used to inspect the complete safetensors headers and every `.comfy_quant` payload from all five released artifacts without downloading model-weight payloads. Layer counts, dtypes, tensor shapes, group sizes, and the pruned AdaLN table matched the runtime contract.

The released Qwen3-VL INT8 artifact has 552 ordinary BF16 tensors and no ordinary F32 tensor; all 350 F32 entries are ConvRot scales.

## Validation

- MiniMax-H3 training, sampling, and cache tests: `99 passed`
- complete repository test suite: `353 passed`
- Ruff check and format check: passed
- `py_compile` and `git diff --check`: passed
- CUDA block-swap LoRA backward tested with pinned staging both disabled and enabled

The complete suite used a process-local compatibility shim because this machine's global environment has an unrelated `huggingface-hub==1.24.0` / Transformers constraint mismatch and a broken optional `flash_attn` DLL. The global Python installation was not modified.


kohya-ss:
Oh, I'm sorry—I actually just finished implementing INT8 ConvRot support in #1025.

I will check if there is anything I can incorporate from this PR. I am truly sorry.

kohya-ss:
Thank you again for this PR — I've now reviewed it in detail, and it contains real value beyond the overlap with #1025: the INT8 ConvRot text encoder loading, the pruned transformer support (AdaLN curve table), the runtime LoRA attachment that works directly on an INT8 base (so users who only have the INT8 download can still generate with LoRAs), and the header-only artifact validation.

Rather than asking you to rebase — which would be a lot of work since #1025 has just landed in `dev` and rewrote the same core files — I'd like to take over the integration myself, if that's fine with you. Since "Allow edits by maintainers" is enabled, I'll push commits directly onto this branch:

- rebase onto current `dev` and replace the overlapping full-transformer INT8 portion with the #1025 implementation (which additionally supports on-the-fly quantization of BF16 checkpoints, verified bit-exact against the ComfyUI distribution)
- keep and adapt this PR's unique parts on top: text encoder INT8, pruned transformer support, runtime LoRA attachment for INT8 bases, and the artifact inspection/validation
- adopt this PR's auto-detection UX for pre-quantized files (no flag needed); `--convrot_int8` stays for dynamic quantization of BF16 sources
- run real-weight verification on my side (I have the released INT8 artifacts locally)

The PR will stay under your authorship. I'll comment again when the branch is updated. Thank you for the thorough validation work on the released artifacts — the topology and `comfy_quant` checks were genuinely useful during review.

sdbds:
> Thank you again for this PR — I've now reviewed it in detail, and it contains real value beyond the overlap with #1025: the INT8 ConvRot text encoder loading, the pruned transformer support (AdaLN curve table), the runtime LoRA attachment that works directly on an INT8 base (so users who only have the INT8 download can still generate with LoRAs), and the header-only artifact validation.
> 
> Rather than asking you to rebase — which would be a lot of work since #1025 has just landed in `dev` and rewrote the same core files — I'd like to take over the integration myself, if that's fine with you. Since "Allow edits by maintainers" is enabled, I'll push commits directly onto this branch:
> 
> * rebase onto current `dev` and replace the overlapping full-transformer INT8 portion with the [Add ConvRot INT8 support for MiniMax-H3 (training + generation, ComfyUI pre-quantized loading) #1025](https://github.com/kohya-ss/musubi-tuner/pull/1025) implementation (which additionally supports on-the-fly quantization of BF16 checkpoints, verified bit-exact against the ComfyUI distribution)
> * keep and adapt this PR's unique parts on top: text encoder INT8, pruned transformer support, runtime LoRA attachment for INT8 bases, and the artifact inspection/validation
> * adopt this PR's auto-detection UX for pre-quantized files (no flag needed); `--convrot_int8` stays for dynamic quantization of BF16 sources
> * run real-weight verification on my side (I have the released INT8 artifacts locally)
> 
> The PR will stay under your authorship. I'll comment again when the branch is updated. Thank you for the thorough validation work on the released artifacts — the topology and `comfy_quant` checks were genuinely useful during review.

No problem. I was just trying to test it locally and noticed there weren't any changes to the int8 convrot, so I wrote one. It would be great if this helps speed up the merge.

kohya-ss:
The branch is now updated (c5394fb → a5b90b2), with your commits kept intact. What changed:

- Merged current `dev` (where #1025 landed). The full-transformer INT8 portion now uses that implementation (on-the-fly BF16 quantization included, bit-exact against the ComfyUI distribution and smoke-tested on real weights).
- Your unique contributions are re-integrated on top and are the heart of this PR now: the ConvRot INT8 text encoder, pruned transformer support (curve-table interpolation), runtime LoRA attachment on pre-quantized bases (this replaces #1025's explicit error — users who only have the INT8 download can now generate with LoRAs), and the header-only artifact inspection.
- Two deliberate design changes from your original: `--convrot_int8` stays as the switch for dynamic quantization of BF16 sources (pre-quantized files auto-detect exactly as you designed), and the exact released-topology pinning (250/200/350 layers) is relaxed to per-layer validation plus strict loading, so artifacts with a different quantized layer set load as declared.
- The planning documents under `docs/superpowers/` are dropped from the PR.
- Full suite: 365 passed + ruff clean. Auto-detection verified against the released artifact headers: full transformer 250 layers (group size 64×50 + 256), text encoder 350 layers (group size 256) — matching your HTTP-range validation.

Remaining on my side before merge: a real-weight smoke of the text encoder path and an A/B check of the pruned transformer. I'll report the results here. Thank you again — the artifact validation and the pruned/TE groundwork made this integration much easier.

kohya-ss:
Real-weight verification of the updated branch is complete — all gates passed (single 96 GB GPU, t2va 384x224x39, seed-fixed 10-step generation, decoded diffs in 8-bit units via frame/audio comparison):

**Text encoder INT8** (auto-detected, 350 layers / group size 256, matching your artifact validation):
- TE caching now runs on the GPU (~26.2 GB): 4.5 s/record vs 15.4 s/record for the BF16 encoder on CPU.
- End-to-end A/B against the BF16 TE (same INT8 DiT, same seed): decoded video mean abs diff **3.5/255**, visually indistinguishable.

**Runtime LoRA attachment** (your no-merge route): output differs from the merge-then-quantize route by only **1.0/255** mean — the two routes are interchangeable in practice, and INT8-only users can now generate with LoRAs.

**Pruned transformer** (auto-detected: 200 INT8 layers + the FP32 [1025, 8] curve table): baseline differs from the full INT8 transformer by **3.7/255** mean, consistent with our offline AdaLN A/B (~7.5/255) — this confirms the curve-table interpolation semantics against the released artifact. LoRA on the pruned base works via runtime attachment (0.97/255 vs the same LoRA on the full base).

**Signal anchor**: the LoRA effect itself measures **79.2/255** mean against the no-LoRA baseline, so all quantization/pruning deltas above sit more than 20x below the signal they need to preserve.

**Flag-less pruned training**: 10 steps at 1.38 s/it (same speed as the full INT8 base), loss matching to three digits, and training-time sampling with the INT8 TE works — which resolves the ~50 GB sampling requirement noted in the docs.

I'm merging this into `dev` now; my detailed follow-up refinements will land as separate PRs on top. Thank you again — the text encoder support, pruned classification, and the runtime attachment design all came from this PR and verified cleanly against the released artifacts.

PR #1027, opened 2026-08-08
Add a fallback loader for MiniMax-H3 checkpoints containing mixed

spawner1145:
tensorwise INT8, ConvRot INT8, and NVFP4/AWQ layers.(for example comfy qwen3 checkpoints)

### DIFF
--- a/docs/minimax_h3.md
+++ b/docs/minimax_h3.md
@@ -25,7 +25,7 @@ Download the following files from [Comfy-Org/MiniMax-H3](https://huggingface.co/
 | Video VAE | `vae/minimax_h3_video_vae_fp16.safetensors` |
 | Audio VAE | `vae/minimax_h3_audio_vae_fp32.safetensors` |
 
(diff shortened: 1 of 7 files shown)


kohya-ss:
Thank you for the PR, and for the detailed work on supporting mixed ComfyUI quantized checkpoints.

After reviewing the implementation, I've decided not to merge this PR. As mentioned in #1018, NVFP4 support for MiniMax-H3 is something I'd like to implement on our side. I've been working on an experimental NVFP4 implementation based on a different approach — W4A4 compute using `torch.nn.functional.scaled_mm` from PyTorch 2.10, following the existing Musubi Tuner conventions (monkey-patching rather than module replacement, and a single loading path). Since the implementation direction differs significantly, maintaining both approaches would be difficult, so I'll close this PR.

That said, this PR was genuinely helpful in two ways:

1. **It made the real user need clear.** Loading pre-quantized ComfyUI artifacts — especially the NVFP4/AWQ Qwen3-VL text encoder — was something I had been unsure about supporting, but I now understand there is actual demand for it. I'll make sure our NVFP4 implementation covers loading these pre-quantized weights, not just on-the-fly quantization.

2. **The format details are valuable.** The nibble ordering, the cuBLAS 128x4 swizzled scale layout, and the AWQ `pre_quant_scale` handling are all useful reference information for our implementation. Treating tensorwise INT8 as the groupsize-1 degenerate case of ConvRot is also an interesting idea.

Thank you again for your contribution and understanding.

PR #1037, opened 2026-08-09
[Fast Fix] MiniMax-H3 FlashAttention dtype mismatch under BF16 autocast

sdbds:
## Summary

- align normalized MiniMax-H3 query and key tensors with the value tensor dtype before dispatching to the attention backend
- add a regression test covering RMSNorm-style FP32 promotion under BF16 execution

## Root cause

Under CUDA BF16 autocast, RMSNorm can produce FP32 query and key tensors while value remains BF16. The MiniMax-H3 attention path forwarded those mixed dtypes directly to FlashAttention 2, which only accepts FP16/BF16 inputs and raised `FlashAttention only support fp16 and bf16 data type` during standalone generation. ConvRot INT8 made the issue visible but was not the cause.

## Validation

- `pytest -q tests/test_minimax_h3_model.py`: 39 passed
- real RTX 4090 CUDA BF16 autocast + FlashAttention 2 probe: passed
- real MiniMax-H3 generation with `--attn_mode flash --convrot_int8`: passed
- Ruff check, Ruff format check, `py_compile`, and `git diff --check`: passed
- full local suite introduced no new failures compared with the exact `upstream/dev` baseline

### DIFF
--- a/src/musubi_tuner/minimax_h3/model.py
+++ b/src/musubi_tuner/minimax_h3/model.py
@@ -350,6 +350,8 @@ class Attention(nn.Module):
         if rotation_table is not None:
             query = _apply_rope_split_half(query, rotation_table)
             key = _apply_rope_split_half(key, rotation_table)
+        query = query.to(value)
+        key = key.to(value)
 
         if self.attn_mode == "flash3":
             try:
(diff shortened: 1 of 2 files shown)


kohya-ss:
Thank you! I use SDPA myself for testing, so I hadn't noticed this.

PR #1038, opened 2026-08-09
Fix MiniMax-H3 activation CPU offload device mismatch

sdbds:
## Summary

- restore MiniMax-H3 checkpointed activations to the transformer execution device before the final layer
- add a CUDA regression test covering activation CPU offload through forward and backward

## Root cause

`create_cpu_offloading_wrapper` returns each checkpointed block output on CPU. After the last block, MiniMax-H3 passed that CPU residual stream directly to a CUDA `final_layer`, so RMSNorm failed with a CPU/CUDA device mismatch.

## Impact

MiniMax-H3 training can use `--gradient_checkpointing_cpu_offload` without failing at the final RMSNorm.

## Validation

- `python -m pytest tests/test_minimax_h3_model.py -q` (40 passed)
- `..\.venv\Scripts\ruff.exe check src/musubi_tuner/minimax_h3/model.py tests/test_minimax_h3_model.py`

### DIFF
--- a/src/musubi_tuner/minimax_h3/model.py
+++ b/src/musubi_tuner/minimax_h3/model.py
@@ -944,6 +944,9 @@ class MiniMaxH3Model(nn.Module):
             if self.blocks_to_swap:
                 self.offloader.submit_move_blocks_forward(self.blocks, index)
 
+        if hidden_states.device != execution_device:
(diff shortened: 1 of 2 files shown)


kohya-ss:
Thank you for this!

PR #1044, opened 2026-08-10
Add MiniMax-H3 image training

sdbds:
This carries the image-training work reviewed in [sdbds/musubi-tuner#2](https://github.com/sdbds/musubi-tuner/pull/2) to upstream `dev`.

## Summary

- Add MiniMax-H3 FL2VA image training modes for first-frame and first/last-frame conditioning.
- Support still-image and multi-frame targets with per-dataset H3-compatible `17*k+5` frame counts.
- Support unsuffixed, zero-indexed, one-indexed, and zero-padded target filenames while keeping H3 continuity checks out of other architectures.
- Validate JSONL target/control indexes and resolve relative media paths from the JSONL directory.
- Build cache paths from structured item fields rather than reverse-parsing filenames.
- Add an ordered sample fingerprint covering target/control roles, media identities, frame count, bucket geometry, and resize policy.
- Require paired latent/text image caches to carry the same sample fingerprint.
- Prepare text-cache image samples from structured dataset metadata without reloading target pixels.
- Share H3 image-mode constants and text-visual pixel limits between cache and generation paths.
- Warn when `--h3_select_frame` is clamped; image-file output no longer requires the unused audio VAE checkpoint.
- Keep image-mode experimental-duration handling explicit and reject invalid frame counts instead of rounding them.
- Update MiniMax-H3 documentation and regression coverage.

## Upstream synchronization

The branch includes `upstream/dev@a892a04`, including the MiniMax-H3 guidance-distillation update from #1045. It uses the current upstream MiniMax-H3 quantization implementation exclusively; the earlier PR-local ConvRot/Comfy quantization code is not included.

## Verification

- `ruff check .`
- `ruff format --check` for the H3 and touched shared files
- `python -m pytest -q`
- Result: **453 passed, 44 existing deprecation warnings**

## Note

Image training uses dummy, unsupervised target audio. Since MiniMax-H3 is a joint video/audio transformer, image-only LoRAs should be treated as unconstrained for audio output. Existing image caches created before the common sample fingerprint should be rebuilt on both latent and text sides.


kohya-ss:
I haven't fully reviewed it yet, but the direction might differ slightly from what I am considering for the image training—for instance, regarding the support policy for multi-frame targets or the method for extending dataset classes.

Also, beyond training LoRA as an image editing model, I would like to train a LoRA that can be applied to video generation.

Therefore, please be aware that changes may be made after the merge.

kohya-ss:
Thank you for the substantial work here — the test coverage is thorough, and several pieces of
this PR are genuinely valuable (more on that below). After careful review I've decided not to
merge this PR, and I'd like to explain the technical reasoning in detail, because it comes from
experiments rather than style preferences.

## Why not merge

**The conditioning scheme conflicts with the base model's strongest prior.** In `first` /
`first_last` training, the FL2VA anchors carry the *input* image at the same RoPE timestamps
where the target rows carry the *output* image. We recently measured how the released model
treats timestamp-coinciding anchors: placing a 1-token target at exactly the last anchor's time
reproduces that anchor almost pixel-exactly (RMS ≈ 0.012 in [0,1] space). Anchor-copying at
matching timestamps appears to be one of the strongest behaviors in the model. Training data
that puts the input at the target's own start/end timestamps while supervising the output there
asks the LoRA to unlearn that behavior globally — which we expect to cost edit quality,
training time, and possibly video behavior of the same LoRA.

The direction we want for H3 image training instead separates `target_index` from
`control_index` on the 24 fps frame grid (the same shape as Wan's one-frame training), so the
conditions keep their natural timestamps and the target sits elsewhere. We also measured that a
true `T_lat=1` target works at inference (with a duplicate-to-2-tokens decode fix for the
single-token VAE path), which halves the target rows relative to the 5-frame grid. A design doc
for this "one-frame mode" (t2va / first-only / first+last with explicit indices) will go to the
H3 roadmap issue; your `first_last` use case maps onto it as the two-condition special case,
and collaboration there is very welcome.

Because the mode API (`--h3_image_mode`, `h3_image_frame_count`, and the cache contracts built
around it) is tightly coupled to the anchors-at-target-timestamps scheme, converting this PR to
the index-based design would amount to a rewrite, so closing is cleaner than merging and then
deprecating the surface.

## What we will adopt (with credit)

Three pieces of this PR are good ideas independent of the conditioning scheme, and I'd like to
bring them in separately, crediting this PR:

1. The **shared sample fingerprint** tying a latent/text cache pair together. Since FL2VA text
   presentations bake condition images in, pair integrity checking is the right call.
2. The **cache-path bookkeeping fix** in the shared cache scripts (tracking final paths after
   an encoder rewrites them) — a real latent bug today.
3. The **text-encoder visual pixel cap** for large images passed to Qwen3-VL.

The datasource hardening also contains good changes, but it alters behavior for all
architectures (e.g. JSONL relative-path resolution), so we'll take only the minimal parts we
need, separately.

Thank you again — the review of this PR directly sharpened the one-frame design, and I hope
you'll join the discussion on the roadmap issue.

PR #1047, opened 2026-08-11
Add teacher-matching training for MiniMax-H3 T2VA LoRA (endpoint and self-reference teachers)

kohya-ss:
## Summary

Adds `--h3_teacher_matching` to `minimax_h3_train_network.py`: a T2VA LoRA is trained against the frozen base model's predictions under privileged conditioning (generalized distillation / LUPI), instead of the flow-matching target. The teacher shares weights with the student: the same transformer runs once per step as a no-grad forward with the LoRA disabled, conditioned on information from the training clip that the text-only student never sees.

Two teacher condition sets are available through `--h3_teacher_conditions`:

- **`first,last`** (default): FL2VA teacher conditioned on the real first/last frames of the clip.
- **`ref`**: Ref2VA teacher conditioned on the training clip itself — the cached target video+audio latents as the copy-source reference, plus the official editing-style copy declaration (`fully_preserved` / `fully_copy`) wrapped around the caption automatically at cache time. This gives the teacher complete information at every sigma, and because the reference audio is copied, **audio becomes a real teaching target** in this mode (with `first,last` it degenerates to a base-preservation anchor).

The teacher prediction lives in the distilled guided space, so no guidance scale and no uncond probe are needed, and the de-distillation drift of plain flow targets is structurally avoided — this loss is mutually exclusive with `--h3_guidance_loss_scale` (#1045).

## Usage

Endpoint (FL2VA) teacher:

- Latent caching: `--task fl2va` (existing format; the first/last conditions feed only the teacher)
- Text caching: `--task t2va --teacher_conditions first,last`
- Training: `--task t2va --h3_teacher_matching`

Self-reference (Ref2VA) teacher:

- Latent caching: unchanged — existing `--task fl2va` caches work as-is (the endpoint latents are simply unused) and plain `--task t2va` caches suffice; the reference condition latents at training time are the cached target latents themselves
- Text caching: `--task t2va --teacher_conditions ref`
- Training: `--task t2va --h3_teacher_matching --h3_teacher_conditions ref`
- The two teacher kinds use distinct cache tensor keys, so a cache/flag mode mismatch is a hard error rather than a silent layout desync

Validated starting recipe for identity training (character data):

```
--h3_teacher_matching --h3_teacher_loss_dc_weight 0.3 --h3_timestep_focus_prob 0.5
```

## What's included

Training:

- **Teacher-matching loss** with the two teacher kinds above; the `--h3_teacher_conditions` seam reserves the interface for further variants (single-sided, anchored, segmented teachers)
- **`--h3_teacher_condition_sigma_max`** (default 0.75): above this drawn base sigma the teacher drops its conditions and runs on the student's own text — a pure base-preservation anchor. The teacher target is a noiseless regression label, so unrestricted teaching near pure noise rapidly overwrites the base composition prior with the dataset mean; the anchor also counters collateral drift from the LoRA's shared weights. For the ref teacher the same band is where the FL2VA weights fail to align the reference against a noise-dominated `x_t` (measured, structural — prompting does not fix it), so the gate boundary coincides with the edge of the healthy range
- **Decomposed loss** (the standard teacher-matching loss): the exact magnitude/direction split of the MSE with the norm-shrinkage coupling removed. Plain MSE rewards hedging an unpredictable direction by shrinking the norm, driving the student to the conditional mean's reduced magnitude — seen at inference as delayed content commitment and washed-out contrast. At default weights the loss value still equals the MSE (only the gradient geometry changes); `--h3_teacher_loss_mag_weight` lowers the magnitude term relative to the direction term
- **`--h3_teacher_loss_dc_weight`**: attenuates the video residual's per-channel DC component (a global palette/tone cast, the style axis) on conditioned teaching steps only — measured at ~7% of the teaching-band residual energy but fully coherent, so it accumulates into a visible style shift. Preservation-anchor steps and the audio anchor keep their full DC penalty: there it is what pulls palette drift back to the base
- **`--h3_teacher_preservation_weight`**: anchor-step loss weight, applied on top of an automatic density compensation that keeps the anchor's expected gradient share invariant under timestep focus
- **`--h3_timestep_focus_min/max/prob`** (any H3 training, not only teacher matching): mixture sampling that concentrates the base-sigma draw on an interior band (default 0.4–0.8, the content-decision band) while the rest of [0,1) keeps `(1-prob)` coverage — the shift family can only pile mass onto an endpoint, not an interior band
- **Fix: `set_enabled` was a no-op for training LoRA/LoHa/LoKr modules** — only the inference module classes honored `enabled`, so the teacher forward initially ran with the LoRA still active (confirmed in smoke logs: a frozen teacher's flow gap must be time-invariant but drifted +30% in 100 steps). Fixed at the training module base classes, with real-network regression tests

Inference (`minimax_h3_generate_video.py`):

- **`--lora_runtime_attach`**: the destructive BF16 merge rounds fused weights to the BF16 mantissa grid, which attenuates or erases small-magnitude adapters; teacher-matching LoRAs equilibrate at exactly such magnitudes (measured: only ~28% of merged delta elements survive rounding). The flag forces the runtime additive-branch route (already used for pre-quantized INT8 bases) on any base, reproducing the training-time forward exactly. Useful beyond teacher matching for any small-amplitude adapter
- **`--trajectory_dir`**: decodes the per-step clean estimate `x0_hat = x_t + sigma*v` to one mp4 per sampling step plus a `sigma_schedule.csv` — shows at which sigma content and style commit; used here to locate the identity-decision band (base sigma 0.6–0.75)

Observability: per-step `teacher/base_sigma`, `teacher/conditioned`, `teacher/{video,audio}_flow_gap_rms`, direction/magnitude decomposition (`teacher/{video,audio}_cos`, `teacher/{video,audio}_norm_ratio`), and the residual DC/AC split (`teacher/{video,audio}_residual_{dc,ac}_rms`) for sigma-binned analysis.

Docs section in `docs/minimax_h3.md` (including a Reference teacher subsection with a starting recipe and monitoring guidance), metadata (`ss_minimax_h3_teacher_*`), and tests (all 464 pass).

## Measured results (real hardware, character data)

Endpoint teacher (500-step reduced-resolution run): the student-teacher residual declines exactly where the teacher's privileged information lives (base sigma 0.6–0.75: loss −59%, cos 0.877→0.939) with low-sigma bands flat; timestep focus roughly doubles wall-band convergence; the decomposed loss holds the wall-band norm ratio at ~1.0 (the washout mechanism is removed — confirmed visually at inference).

Ref teacher (100-step run): the teaching-band flow gap collapses to the probe-predicted flat floor (rms 0.11–0.13 across base sigma 0.15–0.75, vs 0.19–0.31 for the endpoint teacher — 2.5x more teacher precision in the identity-decision band); audio education opens up (teacher audio gap 0.10–0.23 vs 0.46–0.72, and the audio loss becomes a real learning signal instead of an anchor floor); the preservation-anchor band is unchanged; learnable signal appears in the low-sigma bands the endpoint teacher could not serve. Sample quality visibly improves over the endpoint teacher at the same step count.

Ref teacher (500-step runs, including a paired `--audio_loss_weight` 1.0-vs-0.25 A/B on identical noise/timestep sequences): all mechanisms stay healthy over the long run — the preservation anchor reaches an approximate equilibrium (clearly milder drift than the pre-DC-gate-fix endpoint run), and there is no de-distillation signature (education-band norm ratios hold above 1). The teaching-band residual plateaus around step 300, and the strongest checkpoints sit at or just after the plateau; training past it mostly optimizes irreducible floors while the anchor-band drift keeps growing, so sample quality oscillates with the drift-and-recover cycle. The A/B localized the audio term's role: although it carries over half of the education loss and is largely unlearnable on audio-inconsistent data, it is essentially neutral for video (education unchanged; anchor drift marginally larger at the *lower* weight via reduced Adam second-moment damping) — the video-quality levers are the step budget, `--h3_teacher_preservation_weight`, and the LR schedule, and `--audio_loss_weight` is an audio-policy choice. All of this is documented in the ref-teacher caveats.

## Experimental status (why draft)

The interface is considered stable — anticipated follow-ups (further teacher variants, recipe knobs) slot in additively, and cached text staleness is fingerprint-managed. The 500-step health gate has passed (anchor equilibrium, no de-distillation, strong at-plateau checkpoints, audio-weight A/B resolved). Remaining work is recipe tuning that can continue on `dev` after merge:

- caption design guidance (detailed appearance descriptions let the base predict the character from text alone, shrinking the learnable residual; trigger-word-centric captions are the counter-experiment)
- final per-task recommendations for `--h3_teacher_condition_sigma_max` and the focus band (the ref teacher unties the learning signal from high sigma, making a lower focus band a live option)
- long-run management (preservation weight and LR-decay recipes past the teaching-band plateau)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rockerBOO:
Is there a related paper or other context on this approach? Hard to understand this approach from the PR.

kohya-ss:
As far as I know, this is a new idea. I propose using the difference between the inference results of the t2va model (with LoRA) and the fl2va model (using first/last frames but without LoRA, target) as the loss. This should train the LoRA to learn the target video without de-distillation.

The document provides some explanation: https://github.com/kohya-ss/musubi-tuner/blob/feat-h3-teacher-matching/docs/minimax_h3.md#teacher-matching-training-fl2va-teacher-for-a-t2va-student

PR #1051, opened 2026-08-12
fix: pin ruff below 0.16.0 to keep CI lint passing

rockerBOO:
## Summary

ruff 0.16.0 enabled a large batch of new lint rules by default. With no changes to our code, upgrading to it produces ~2000 new findings across the repo (CI run: https://github.com/kohya-ss/musubi-tuner/actions/runs/31637928085). This pins the dev dependency below 0.16 so CI stays green until those rules are triaged and addressed deliberately.

https://astral.sh/blog/ruff-v0.16.0

## Test plan

- [x] `uv lock` resolves `ruff` to `0.15.22`
- [x] `ruff check` passes clean on the tracked repo with the pinned version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,9 @@ dev = [
   "matplotlib==3.10.0",
   "tensorboard",
   "prompt-toolkit==3.0.51",
-  "ruff>=0.12.10",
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for this! I also noticed this while working on the `dev` branch. While we'll address the issue separately, I think pinning the version is the best immediate measure.

PR #1075, opened 2026-08-30
Support the guidance loss under --task ref2va

christopher5106:
Training with `--task ref2va` and `--h3_guidance_loss_scale` fails on the first step:

```
File ".../minimax_h3/packing.py", line 325, in build_h3_layout
    raise ValueError("MiniMax-H3 condition roles apply only to FL2VA layouts")
```

`_apply_guidance_loss_targets` harvests `condition_roles` from every `visual_condition` segment of the runtime layout, but Ref2VA reference blocks share that segment kind (`ref_000_image`, `ref_001_video`, ...), so the uncond rebuild passes reference roles into `build_h3_layout`, which rejects roles for non-FL2VA tasks. The roles are only needed for FL2VA (the K=1 one-frame case); ref2va reference segments are regenerated deterministically from the layout's `references` tuple.

Fix: harvest the roles for FL2VA layouts only. The uncond probe forward then works for ref2va with the references preserved and only the text rows swapped — the same "matching how the adapted model runs at inference" semantics the docstring already states for visual/audio conditions.

Verified:
- new regression test (mirroring `test_guidance_loss_uncond_layout_carries_one_frame_fl_condition_roles`) fails on current dev and passes with the fix; `tests/test_minimax_h3_training.py` passes in full (138)
- end-to-end: a ref2va LoRA training (5 video targets + image references) that previously crashed at step 0 with this error runs to completion with the guidance loss active (B200, bf16 ref2va checkpoint)

### DIFF
--- a/src/musubi_tuner/minimax_h3_train_network.py
+++ b/src/musubi_tuner/minimax_h3_train_network.py
@@ -1461,9 +1461,16 @@ class MiniMaxH3NetworkTrainer(NetworkTrainer):
             )
         # the probe swaps only the text rows; the one-frame times stay valid because they
         # are relative to the target-block cursor, which moves with the text length. The
-        # condition roles are recovered from the segments so a one-frame FL2VA layout with a
-        # single condition rebuilds identically (roles are required for K=1).
-        uncond_condition_roles = tuple(segment.role for segment in runtime.layout.segments if segment.kind == "visual_condition")
+        # FL2VA condition roles are recovered from the segments so a one-frame FL2VA layout
+        # with a single condition rebuilds identically (roles are required for K=1). Ref2VA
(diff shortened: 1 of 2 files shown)


christopher5106:
Superseded by your own fff5c19 (one-frame Ref2VA training) — closing.

That commit harvests `condition_roles` for FL2VA layouts only and passes `condition_roles=uncond_condition_roles or None`, so the empty tuple clears the "condition roles apply only to FL2VA layouts" guard in `build_h3_layout`. Same effect as this PR, reached while building a larger feature, and `test_guidance_loss_uncond_layout_carries_one_frame_references` covers the path.

Confirmed by merging dev into our fork: the only conflicts were this change against yours, and taking yours leaves guidance loss working under `--task ref2va` for video targets as well as one-frame. Thanks.

PR #1081, opened 2026-09-01
Add NVFP4 pre-quantized training/inference + ConvRot int8 mixed-dtype support for Krea 2

rockerBOO:
This adds NVFP4/convrot int8 support for pre-quantized models and training on the native precision. 

Builds off NVFP4 inference support for MiniMax H3 NVFP4 text encoder support. Kernel is not being used (that I am aware of) for NVFP4 inference but could be changed over to support it. 

On the fly quantization is not currently supported.  

**Related:** loading a checkpoint that mixes NVFP4 and ConvRot INT8 pre-quantized weights *within the
same DiT* (each module's format declared individually) is being developed separately in
[rockerBOO/musubi-tuner#1](https://github.com/rockerBOO/musubi-tuner/pull/1), on top of this branch.

## Requirements

- Triton needed for custom kernels
- PyTorch 2.10 for fp4 supported operators 
- Blackwell chip (50xx, RTX xxxx PRO, B200, B300) 

## Example model 

[rockerBOO/krea2-raw-nvfp4](https://huggingface.co/rockerBOO/krea2-raw-nvfp4). NVFP4 and mixed NVFP4 + ConvRot INT8 models

- https://huggingface.co/rockerBOO/krea2-raw-nvfp4/blob/main/krea2_raw_nvfp4.safetensors
- https://huggingface.co/rockerBOO/krea2-raw-nvfp4/blob/main/krea2_raw_nvfp4_convrot_int8.safetensors

## Kernel / LoRA training

- Adds Triton kernel but supports LoRA training specifically. Need DGRAD and WGRAD properly if doing full model training.

fig_training_flow_linear_layer

[Pretraining Large Language Models with NVFP4](https://arxiv.org/html/2509.25149v1)

Krea 2 specifically supported so we can validate the approach before moving to other models. 

## Performance

Goal is to meet or exceed performance of ConvRot INT8 on nvfp4 based tensors. Benchmarks with nvfp4 kernel compared to ConvRot INT8.

```
mlp.gate [16384,6144]
├─ NVFP4 (fused kernel):  1.47ms / 276 MiB  ⭐
├─ ConvRot INT8:          2.80ms / 568 MiB
└─ Winner: NVFP4 (~1.9x faster, ~2x less memory)

attn.wk [1536,6144]
├─ NVFP4 (fused kernel):  0.35ms / 133 MiB  ⭐
├─ ConvRot INT8:          0.50ms / 191 MiB
└─ Winner: NVFP4 (~1.4x faster, ~1.4x less memory)
```

Current overall performance is slightly slower compared to ConvRot INT8. 5.5s/it compared to 6s/it. Also memory usage is higher overall compared to ConvRot INT8. 14GB compared to 11GB, so some further tuning is required. 

## Summary
- Add true FP4x4 (NVFP4) tensor-core forward/backward training against ComfyUI pre-quantized Krea 2 checkpoints (`--nvfp4`), including columnwise backward-weight requantization, block-swap streaming of NVFP4 buffers, stochastic-rounding activation quantization, and a fused Triton NVFP4 activation-quantization kernel.
- Add standalone inference support: `krea2_generate_image.py` now accepts `--nvfp4`/`--convrot_int8` to sample directly from a pre-quantized checkpoint, without going through the trainer's sample-image loop. Fixes a hard blocker where `load_krea2_dit` always built training-only NVFP4 backward buffers regardless of caller.
- Share quantization-scheme validation (mutual exclusivity, NVFP4 runtime requirements) between the trainer and the new inference script via `modules/quantization_utils.py` and `krea2_utils.validate_krea2_quantization_args`, instead of duplicating checks.
- Extensive test coverage: shape/roundtrip edge cases, training-vs-inference forward-path/buffer boundaries, block-swap selector auto-detection, stochastic-rounding unbiasedness, CLI-level validation.

## Training test plan 
Step 1: Run a handful of real training steps
```
uv run --no-sync accelerate launch --num_cpu_threads_per_process 1 --mixed_precision bf16 \
  src/musubi_tuner/krea2_train_network.py \
  --dit krea2_raw_nvfp4.safetensors \
  --nvfp4 \
  --network_module networks.lora_krea2 \
  --network_dim 16 \
  --max_train_steps 10 \
  --dataset_config <a small existing krea2 dataset config> \
  --output_dir /tmp/nvfp4-smoke-test \
  --output_name nvfp4_smoke \
  --logging_dir /tmp/nvfp4-smoke-test/logs \
  --log_grad_metrics
```

Step 2: Verify pass criteria
- Loss finite for all 10 steps, not obviously diverging
- --log_grad_metrics's pre-clip gradient norm finite (no NaN/Inf)
- No exceptions during load, forward, backward, or optimizer step

## Inference test plan
- [x] `uv run --no-sync --with pytest pytest tests/test_quantization_utils.py tests/test_krea2_nvfp4.py tests/test_krea2_convrot_int8.py tests/test_krea2_generate_image_quantization.py tests/test_nvfp4_training.py` — 88 passed
- [x] Manual GPU smoke test on real Blackwell hardware with a real NVFP4 checkpoint (standalone inference, with/without block swap, LoRA-rejection path
- [x] Basic standalone inference — run krea2_generate_image.py --nvfp4 against a real checkpoint; confirm it completes and logs (nvfp4) with no LoRA-merge suffix.
- [x] Block swap, both offloader modes — run with --nvfp4 --blocks_to_swap 8, confirm it works without requiring --block_swap_h2d_only (this is the design's key claim — that the training-only requirement doesn't apply at inference). Then repeat with --block_swap_h2d_only --use_pinned_memory_for_block_swap added, and confirm both produce visually equivalent output.
- [x] LoRA rejection path — run with --nvfp4 --lora_weight <file>, confirm it fails fast with the ValueError from main() before any model loading, not a deep stack trace from inside load_krea2_dit.


kohya-ss:
Thank you for the extensive work here. A quick status note: `dev` is about to be merged into `main` (MiniMax-H3 and Krea 2 ConvRot int8), and the trainer / offloader / quantization-utils code moved quite a bit in the process, so this branch currently conflicts with `dev`.

Rather than review it against a moving target, I would like to look at it after the `main` merge lands. Could you rebase onto `dev` (or `main`) at that point? Since this stays a draft until the Blackwell smoke tests in your test plan are checked off, there is no rush from my side. I also have a Blackwell environment available, so please let me know if there's anything I can do to help.

rockerBOO:
Got it, I will update for these changes. The test plan should be updated as it all works. Maybe with all the PR's listed below to fix some edge cases. 

~~https://github.com/rockerBOO/musubi-tuner/pull/1 - Load mixed NVFP4 + ConvRot INT8 prequantized Krea 2 checkpoints~~
https://github.com/rockerBOO/musubi-tuner/pull/2 - Flux.2 NVFP4 + ConvRot INT8 training support
https://github.com/rockerBOO/musubi-tuner/pull/4 - fix: ConvRot INT8 + plain block swap (ModelOffloader) crash
https://github.com/rockerBOO/musubi-tuner/pull/5 - fix: keep blocks resident when one backward walks several forward graphs
https://github.com/rockerBOO/musubi-tuner/pull/9 - fix(minimax-h3): support pure NVFP4 transformer checkpoints

I have a stack of 5 more PR's but they all work towards the NVFP4 and ConvRot INT8 preload/mixed training. 

My goal was I need to get more efficiency to do some experimental training like GRPO, Self-Flow which needs a bit of everything to work on something like Krea 2. And additionally got Minimax H3 training on 16GB with a NVFP4 pre-quantized model. 

Additionally, by supporting pre-quantized, the loading would be a lot faster and the quantization could be more accurate.

rockerBOO:
Went through another correctness pass for the nvfp4 kernels and training. 

Merged in mixed quantizer pass which is mostly the same as this PR.
- https://github.com/rockerBOO/musubi-tuner/pull/1 Load mixed NVFP4 + ConvRot INT8 prequantized Krea 2 checkpoints. 

Biggest things to look for 

- MixedQuantizer approach. Maybe a different way we want to do this? 
- NVFP4 with Triton and without could be working but I haven't tested it enough without Triton and it's a lot slower.
- NVFP4 has no onload quantizing path (like --fp8 scaled and convrot int8) and not sure if that is a good idea or not. 
- We are setting --convrot_int8 and --nvfp4 to load the tensors from the weights but that also conflicts in it's path with quantizing on load of bf16 to  ConvRot INT8. Might not be clear what might happen depending on the model. Since we are just loading from the tensors, maybe we don't need to pass either of these and just load it as it is in the model files. 
- We are assuming comfy format in the pre-quantized models but could be different formats for handling the extra tensor keys.
- Scaled FP8 integration is specifically missing if a weight has scaled fp8 it might be weird. Also considering how the current scaled FP8 could be intergrated into this approach for pre-scaled weights. 
- Does things still work outside a blackwell card for other things? Meaning that the NVFP4 implementation is properly isolated. 
- Performance of NVFP4 fully should be nearer to 2x faster than ConvRot INT8, and is close in my experience. 
- Memory usage should also be lower than ConvRot INT8 but haven't looked into it as deeply to compare. 

## Krea 2 — minimal test run (mixed NVFP4 + ConvRot INT8)

Smoke test for `krea2_train_network.py`: loads a ComfyUI checkpoint whose Linears
mix NVFP4 and ConvRot INT8 (`--nvfp4 --convrot_int8`), with block swap on.
Training only — caching is assumed already done.

### dataset_config.toml

```toml
[general]
resolution = 1024
caption_extension = ".txt"
enable_bucket = true

[[datasets]]
image_directory = "/path/to/images"
cache_directory = "/path/to/cache"
```

### test_krea2.toml

```toml
# --- the feature under test: mixed NVFP4 + ConvRot INT8 checkpoint loading ---
nvfp4 = true
convrot_int8 = true
convrot_int8_bwd = "bf16"

# Exercises the LoRAStreamOffloader ring on a mixed-format checkpoint (the block-swap fix).
blocks_to_swap = 24
block_swap_h2d_only = true
block_swap_ring_size = 2

# --- minimal LoRA ---
network_module = "networks.lora_krea2"
network_dim = 8
network_alpha = 8

# --- flow matching ---
timestep_sampling = "krea2_shift"
weighting_scheme = "none"

# --- optimizer ---
optimizer_type = "AdamW"
optimizer_args = ["weight_decay=0.1"]
learning_rate = 1e-4
max_grad_norm = 1.0

# --- smoke test ---
max_train_steps = 20
save_every_n_steps = 20
save_precision = "bf16"
seed = 42
gradient_checkpointing = true
sdpa = true

sample_every_n_steps = 20
sample_at_first = true
```

### sample_prompts.txt
```
a retro anime portrait --w 1024 --h 1024 --s 8 --l 1 --n blurry, low quality, watermark --d 0
```

Turbo LoRA: https://huggingface.co/rockerBOO/krea-2-turbo-lora-r64-kohya/blob/main/krea2_turbo_lora_rank_64_bf16_kohya.safetensors

### Command

```bash
accelerate launch --num_cpu_threads_per_process 1 --mixed_precision bf16 \
  src/musubi_tuner/krea2_train_network.py \
  --text_encoder /path/to/qwen3vl_4b_bf16.safetensors \
  --dit /path/to/krea2_raw_nvfp4_convrot_int8.safetensors \
  --vae /path/to/qwen_image_vae.safetensors \
  --turbo_lora krea2_turbo_lora_rank_64_bf16_kohya.safetensors \
  --dataset_config /path/to/dataset_config.toml \
  --config_file /path/to/test_krea2.toml \
  --output_dir /path/to/out \
  --output_name krea2-smoke \
  --sample_prompts /path/to/sample_prompts.txt \
  --mixed_precision bf16
```

PR #1092, opened 2026-09-06
Fix --attn_mode sdpa not working in shared attention backends

### LINKED ISSUE #1077: --attn_mode sdpa is offered by the CLIs but unimplemented ("Unsupported attention mode: sdpa" / KeyError)
## Summary

Every generation CLI advertises `sdpa` in its `--attn_mode` choices:

```
--attn_mode {flash,torch,sageattn,xformers,sdpa}
```

but neither dispatcher implements it, so passing the documented value fails:

- `modules/attention.py` → `ValueError: Unsupported attention mode: sdpa` (e.g. `flux_2_generate_image.py`)
- `hunyuan_model/attention.py` → `KeyError: 'sdpa'` at `MEMORY_LAYOUT[mode]` (e.g. `qwen_image_generate_image.py`)

This is easy to trip over because the *training* scripts use a `--sdpa` flag for the same backend, so `--attn_mode sdpa` looks like the natural spelling when moving from training to inference.

## Cause

Both dispatchers implement the PyTorch scaled-dot-product path under the name `torch`; `sdpa` was added to the argparse choices without a corresponding branch.

## Fix

Either drop `sdpa` from the choices, or accept it as an alias for `torch`. The alias seems friendlier — the two names mean the same thing to users. Minimal version:

```diff
# modules/attention.py, AttentionParams (dataclass)
+    def __post_init__(self):
+        if self.attn_mode == "sdpa":
+            self.attn_mode = "torch"
```

```diff
# hunyuan_model/attention.py
+MEMORY_LAYOUT["sdpa"] = MEMORY_LAYOUT["torch"]
```
plus normalising `mode` at the top of `attention()` in that module.

Verified on musubi-tuner v0.3.4 with `qwen_image_generate_image.py` and `flux_2_generate_image.py` (torch 2.11 + ROCm 7.2.3, gfx1201): with the alias, `--attn_mode sdpa` behaves identically to `--attn_mode torch`. Happy to send a PR.

rossnot:
--attn_mode sdpa is listed as a valid argparse choice but has no implementation branch in either dispatcher — modules/attention.py raises "Unsupported attention mode: sdpa" and hunyuan_model/attention.py hits KeyError: 'sdpa'. The underlying implementation is PyTorch's scaled-dot-product attention, already present as "torch".

Adds "sdpa" as an alias for "torch" at both entry points: AttentionParams.__post_init__ (covers all shared-attention backends) and the top of hunyuan_model/attention.attention() (covers the Hunyuan path before the MEMORY_LAYOUT lookup).

Fixes #1077.

### DIFF
--- a/src/musubi_tuner/hunyuan_model/attention.py
+++ b/src/musubi_tuner/hunyuan_model/attention.py
@@ -127,6 +127,8 @@ def attention(
     q, k, v = q_or_qkv_list if type(q_or_qkv_list) == list else (q_or_qkv_list, k, v)
     if type(q_or_qkv_list) == list:
         q_or_qkv_list.clear()
+    if mode == "sdpa":
+        mode = "torch"
     split_attn = total_len is not None
     if (split_attn or cu_seqlens_q is None) and mode == "sageattn":
         mode = "sageattn_fixlen"
--- a/src/musubi_tuner/modules/attention.py
+++ b/src/musubi_tuner/modules/attention.py
@@ -37,6 +37,10 @@ class AttentionParams:
     cu_seqlens: Optional[torch.Tensor] = None
     max_seqlen: Optional[int] = None
 
+    def __post_init__(self):
+        if self.attn_mode == "sdpa":
+            self.attn_mode = "torch"
+
     @staticmethod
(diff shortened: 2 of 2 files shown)


kohya-ss:
Thank you for this update!

Issue #1060, opened 2026-08-16
MiniMax H3: shared cache_directory leaks validation samples into training buckets


lcdrzxlcdrzx:
### Summary

When a training dataset and a validation dataset use separate JSONL files but the same `cache_directory`, MiniMax H3 training does not honor the JSONL split at the bucket-building stage. It globs every `*_mmh3.safetensors` file in the shared cache directory, causing validation samples to be included in the training dataloader.

### Environment

- MiniMax H3 Ref2VA-Omni LoRA training
- 64 cached videos, split deterministically into 56 training and 8 validation items
- separate `train_56.jsonl` and `val_8.jsonl`
- initially, both TOML files pointed to the same cache directory

### Reproduction

1. Build 64 latent/text cache pairs in one directory.
2. Create two JSONL files: one with 56 video paths and one with the remaining 8 paths.
3. Configure the training TOML with the 56-item JSONL and the shared cache directory.
4. Start training.
5. Log output reports `loaded 56 videos` from the JSONL but then bucket counts and `num train items` equal 64, not 56.

### Root cause

In `src/musubi_tuner/dataset/image_video_dataset.py`, bucket construction enumerates the cache directory with a glob similar to:

```python
latent_cache_files = glob.glob(os.path.join(self.cache_directory, f"*_{self.architecture}.safetensors"))
```

It then builds ItemInfo from every matched cache file. It does not restrict the cache files to sample identifiers supplied by the configured JSONL datasource.

### Verified workaround

Use independent cache directories for train and validation. We moved cache pairs by video stem into:

- train cache: 56 latent + 56 text files
- validation cache: 8 latent + 8 text files

After updating each TOML to its matching cache directory, the log correctly reported:

```text
loaded 56 videos
total batches: 56
num train items: 56
```

### Proposed fix

Construct bucket inputs from the datasource's resolved item list, or filter discovered cache files to the item keys present in the configured JSONL. A trainer should never silently include unrelated cache files just because they exist in the cache directory.

### Why this matters

The current behavior invalidates validation loss: validation samples leak into training even though users supplied separate JSONL files. This is especially harmful for small video datasets, where validation loss is a primary signal for overfitting.

---

kohya-ss:
[...]

PR #1093, opened 2026-09-06
Fix shared cache directory leaking other splits into training buckets

rossnot:
`prepare_for_training()` in both `ImageDataset` and `VideoDataset` discovers training items by globbing the `cache_directory` for `*_{architecture}.safetensors` files, completely ignoring which items the JSONL datasource actually declares. When train and validation configs point to the same `cache_directory` (a common pattern for large datasets where caching is expensive), validation cache files silently contaminate training batches. The datasource logs `loaded 3 videos` while the bucket logs `count: 10`, with no warning.

This builds a `valid_basenames` set from the JSONL datasource's item paths and skips any cache file whose derived basename is not in that set. Directory-based datasources are unaffected (`valid_basenames` stays `None`, loop is unchanged). The baseline case of separate cache directories per split is also unaffected.

Verified on Wan2.1 T2V 1.3B with 10 cached videos and a 3-item train JSONL pointing to the shared cache:

- Before: `loaded 3 videos` / `bucket: (512, 512, 25), count: 10` / `total batches: 10`
- After: `loaded 3 videos` / `bucket: (512, 512, 25), count: 3` / `total batches: 3`

### DIFF
--- a/src/musubi_tuner/dataset/image_video_dataset.py
+++ b/src/musubi_tuner/dataset/image_video_dataset.py
@@ -561,6 +561,16 @@ class ImageDataset(BaseDataset):
         # glob cache files
         latent_cache_files = glob.glob(os.path.join(self.cache_directory, f"*_{self.architecture}.safetensors"))
 
+        # when a JSONL datasource is used, restrict to items it actually declares; without this,
+        # a shared cache directory leaks files from other splits (e.g. validation) into training
+        valid_basenames = None
+        if isinstance(self.datasource, ImageJsonlDatasource):
+            valid_basenames = set()
+            for item in self.datasource.data:
+                path = item.get("image_path") or item.get("image_path_0")
+                if path:
+                    valid_basenames.add(os.path.splitext(os.path.basename(path))[0])
+
         # assign cache files to item info
         # (width, height) -> [ItemInfo] or (width, height, other conds...) -> [ItemInfo]
         bucketed_item_info: dict[Union[tuple[int, int], Any], list[ItemInfo]] = {}
@@ -572,6 +582,8 @@ class ImageDataset(BaseDataset):
             image_size = (image_width, image_height)
 
             item_key = "_".join(tokens[:-2])
+            if valid_basenames is not None and item_key not in valid_basenames:
+                continue
             text_encoder_output_cache_file = os.path.join(self.cache_directory, f"{item_key}_{self.architecture}_te.safetensors")
             if not os.path.exists(text_encoder_output_cache_file):
                 logger.warning(f"Text encoder output cache file not found: {text_encoder_output_cache_file}")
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for the fix and the clear reproduction. I agree with the symptom, but I want to look at the dataset handling as a whole before deciding on the shape of the fix (the leak is one instance of a more general problem with datasets sharing a cache directory). I will come back to this PR after the upcoming `dev` → `main` merge; leaving it open until then.

kohya-ss:
Thanks again for the PR, and sorry for the wait. Before deciding on the shape of the fix, I would like to understand the use case a bit better.

1. `docs/dataset_config.md` states that the cache directory must be different for each dataset. So the situation in the PR (train and validation configs sharing one `cache_directory`) is outside what the docs currently support. Was this a deliberate proposal to relax that rule, or was the rule not on your radar?

2. This repository does not have a validation dataset feature. Where does the validation side in your setup come from? A fork, an external evaluation script, or simply a held-out set that you cached into the same directory?

3. Is the actual workflow "cache once from a large JSONL, then train with a smaller JSONL that points to a subset"? If so, I would treat this PR as a feature ("restrict training items to the ones declared by the JSONL") rather than a bug fix, independent of validation. To be clear, that does not mean I have decided to add such a feature yet.

4. Would a warning be sufficient for your case? The cheapest change on my side is to warn when the number of items found by the datasource and the number of cache files in the bucket differ, and keep the "one cache directory per dataset" rule as is.

5. The PR leaves directory-based datasources unfiltered. Was that intentional?

Depending on the answers, I would either accept this as an opt-in feature (with a small refactor so the filtering lives in the datasource rather than in isinstance branches in `prepare_for_training`) or replace it with the warning.

rossnot:
Thanks for the review! To answer your questions:

1. I'm not personally facing this. I picked it up from #1060 to contribute. I wasn't aware of the one-directory-per-dataset rule, I see it now on `docs/dataset_config.md`

2. The validation dataset is the reporter's own setup. Separate train.jsonl and validation.jsonl pointing at the same cache dir.

3. Agreed, closer to a feature request than a bug fix.

4. A warning sounds right to me, happy to update the PR.

5. Directory-based datasources were unfiltered which was an oversight on my part.


Happy to keep contributing wherever it's useful.

kohya-ss:
Thanks for the answers, that clears it up. Let's go with the warning. Here is what I have in mind:

- Keep the "one cache directory per dataset" rule as is, and drop the filtering.
- In `prepare_for_training()` of both `ImageDataset` and `VideoDataset`, compare `len(self.datasource)` with the number of distinct `item_key`s collected from the cache files. Every datasource implements `__len__`, so no `isinstance` is needed. For videos, count distinct keys rather than files, because one video produces several cache files.
- Put the comparison in a single helper on `BaseDataset` and call it from both classes.
- Skip the check when the datasource is empty (there is nothing to compare against, and "found 0 images/videos" is already logged).
- Warn on any mismatch, in both directions. More cache items than datasource items points to a shared or stale cache directory; fewer points to an incomplete cache. The message should mention that the cache directory must not be shared between datasets (see `docs/dataset_config.md`).
- Optionally, add one sentence to `docs/dataset_config.md` next to the existing rule saying that sharing a cache directory mixes the items of the other dataset into training.

Please also update the PR title and description to match. I will close #1060 once this is merged, with a note that separate cache directories remain the supported setup.

PR #1101, opened 2026-09-09
Krea 2: wire activation CPU offloading into gradient checkpointing

rockerBOO:
## Summary
- Track an `activation_cpu_offloading` flag on Krea 2's `SingleStreamDiT`
- Wrap the checkpointed block forward with a CPU-offloading wrapper when the flag is enabled, moving activations back to the compute device before the final layer
- Add a mutation-sensitive test covering the offload wiring

## Test plan
- [x] `pytest tests/test_krea2_gradient_checkpointing.py -q` passes


kohya-ss:
Thank you for this!

PR #1102, opened 2026-09-09
Add support for Krea 2 Turbo LoRA for sampling

rockerBOO:
Instead of loading a separate turbo DiT model, we can load the Turbo LoRA and have it work in a chain with the training LoRA.

Turbo LoRA loads at the same time we would load the network weights so works with everything downstream, convrot int8, nvfp4, fp8, and torch compile.

Compatible Turbo LoRA https://huggingface.co/rockerBOO/krea-2-turbo-lora-r64-kohya/blob/main/krea2_turbo_lora_rank_64_bf16_kohya.safetensors


rockerBOO:
Superseded by a clean rebase onto dev (this branch carried unrelated commits from an internal feature branch).

PR #1103, opened 2026-09-09
Add support for Krea 2 Turbo LoRA for sampling

rockerBOO:
Instead of loading a separate turbo DiT model, we can load the Turbo LoRA and have it work in a chain with the training LoRA.

Turbo LoRA loads at the same time we would load the network weights so works with everything downstream, convrot int8, nvfp4, fp8, and torch compile.

Compatible Turbo LoRA https://huggingface.co/rockerBOO/krea-2-turbo-lora-r64-kohya/blob/main/krea2_turbo_lora_rank_64_bf16_kohya.safetensors

### DIFF
--- a/docs/krea2.md
+++ b/docs/krea2.md
@@ -292,6 +292,8 @@ A fox in the snow.  --w 1024 --h 1024 --s 8 --l 1 --d 0
 
 > **`--turbo_dit` cannot be combined with `--blocks_to_swap`.** Turbo sampling swaps the base weights in place, which is only safe without the block-swap offloader. If you use block swap, omit `--turbo_dit` and sample on the RAW model instead.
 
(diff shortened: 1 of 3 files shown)


kohya-ss:
Merged, thank you for this! Verified here on real weights: samples on RAW are garbled at 8 steps and become clean with the Turbo LoRA composed, and the enable/disable toggling around each sample pass logs as expected.

One note for future PRs rather than this one: this repository is moving away from `getattr(args, "...", default)` in favor of plain `args.x` with defaults defined once in argparse (the MiniMax-H3 code has already been converted; the other architectures will follow after the next `dev` → `main` merge). No change needed here, I will fold these into that cleanup.

rockerBOO:
> One note for future PRs rather than this one: this repository is moving away from getattr(args, "...", default) in favor of plain args.x with defaults defined once in argparse 

Does this also handle defaults in the toml config paths? I can check.

